PR #2, opened 2023-03-31
Deployment to fly.io

tuler:
Deployment procedure to fly.io and some other changes to make that possible


apps/api/fly.toml
```diff
@@ -0,0 +1,38 @@
...
+
+#[[services.http_checks]]
+#interval = 10000
+#grace_period = "5s"
+#method = "get"
+#path = "/healthz"
+#protocol = "https"
+#restart_limit = 0
+#timeout = 2000
+#tls_skip_verify = false
```
endersonmaia, line 30: do we need this comments?

```suggestion
```

apps/api/fly.toml
```diff
@@ -0,0 +1,38 @@
...
+protocol = "tcp"
+
+[[services.ports]]
+handlers = ["http"]
+port = 80
+force_https = true
+
+[[services.ports]]
+handlers = ["tls", "http"]
+port = 443
```
endersonmaia, line 20: See https://fly.io/docs/reference/configuration/#the-http_service-section


```suggestion
[http_service]
  internal_port = 8080
  force_https = true
```

Consider adding some concurrency configuration.

```suggestion
[http_service]
  internal_port = 8080
  force_https = true
  [http_service.concurrency]
    type = "requests"
    soft_limit = 200
    hard_limit = 250
```

apps/api/README.md
```diff
@@ -49,3 +49,23 @@ Server ready at http://0.0.0.0:3001
...
 
 OpenAPI browser will be available at `http://localhost:3001/docs/`
+
+## Deployment
+
+The sunodo API and postgres database are currently deployed at [fly.io](http://fly.io/).
+As this is a monorepo deployment must be done from the project root directory, and executing the following command:
+
+```shell
+fly deploy --config apps/api/fly.toml --dockerfile ./Dockerfile
```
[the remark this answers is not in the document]
tuler: My idea is to have a single Dockerfile at root and adding one stage for each product.
The problem of putting inside `apps/api` is because the context need to be at the root, so I think it's more clear to have the Dockerfile at the context location.
Using `docker build` you can specify the context dir and dockerfile location individually, but I'm not sure if it's possible to specify that way with `fly deploy`

apps/api/fly.toml
tuler: Eventually I think we should make this check work, but I couldn't make it work.

endersonmaia: so we could remove this and open an issue

PR #4, opened 2023-03-31
feat(sunodo-registry): add fly app

### LINKED ISSUE #3: Deploy registry at fly.io
Requirements:

- [x] should respond at `registry.sunodo.io`
- [x] should authenticate via api.sunodo.io/
- [x] should store images at S3 https://docs.docker.com/registry/storage-drivers/s3/

endersonmaia:
/close #3

### DIFF
--- a/apps/registry/Dockerfile
+++ b/apps/registry/Dockerfile
@@ -1,0 +1,3 @@
+FROM registry:2.8.1
+
+ADD https://raw.githubusercontent.com/sunodo/sunodo/main/apps/api/keys/registry.crt /keys/registry.crt
--- a/apps/registry/fly.toml
+++ b/apps/registry/fly.toml
@@ -1,0 +1,43 @@
+app = "sunodo-registry"
+primary_region = "ord"
+
+[build]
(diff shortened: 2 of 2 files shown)


apps/registry/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM registry:2.8.1
+
+ADD https://raw.githubusercontent.com/sunodo/sunodo/main/apps/api/keys/registry.crt /keys/registry.crt
```
tuler, line 3: Do you want to add this to the global Dockerfile at root, as a `registry` target?
Then the context is at root and you can copy the keys from the project.

I know this is probably temporary

endersonmaia: I prefer smaller self-container `Dockerfiles`, but I can do that if you think we should go the "single Dockerfile with a lot of targets" path.

tuler: I just don't like adding the file from GitHub as you did

tuler (approved):

PR #12, opened 2023-04-13
ci: add markdownlint

endersonmaia:


oap75 (commented):
We have conflicts !

endersonmaia:
yeah, and we're ganna have them for a while

`oclif` generates this docs, and lint will always fail, unless we find a way to control how `oflic` generate them so that we can adhere to the lining

it's this, or ignoring the linting for the `oflic` generated files for the specific problems it generates

tuler:
Does the linter has a way to format as well? We can run the formatter after the task that generates the README

endersonmaia:
There's `markdownlint-cli2-fix`, just need to test if it's smart enough to fix the cases we need.

tuler:
rebased

tuler:
Not sure if this is workable. The CHANGELOGs generated by changeset are not passing the CI

endersonmaia:
> Not sure if this is workable. The CHANGELOGs generated by changeset are not passing the CI

We could work on exceptions for `MD030` rule for CHANGELOGS, but unless we have other content that uses markdown (docs?), it may not be worth the hassle.

tuler:
> We could work on exceptions for `MD030` rule for CHANGELOGS, but unless we have other content that uses markdown (docs?), it may not be worth the hassle.

Maybe we could exclude the CHANGELOGs from linting. The changeset files can still be linted, as well as the docs.

PR #21, opened 2023-04-20
sunodo build (initial version)

tuler:
This adds:

- a `build` command for the cli
- a `clean` command for the cli
- a `shell` command for the cli


packages/machine-emulator-tools/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+While `cartesi` doesn't provide such image, Sunodo will provide the image under the `ghcr.io/sunodo` organization.
+
+So the application should include the following:
+
+```dockerfile
+COPY --from=ghcr.io/sunodo/machine-emulator-tools:0.10.0-ubuntu22.04 / /
+```
+
+Only `ubuntu 22.04` is currently supported. We might add other distributions in the future.
```
endersonmaia, line 21: We could start with `debian:bullseye` even if wee need to compile server-manager on a sunodo's fork while it's not upstream.

The plan is to run every service from cartesi-node AND server-manager on the same VM, for that, we need it to follow `debian:bullseye`, since that's what's being used on cartesi-node build from the cartei/rollups repository.

tuler: The tools is another issue. This goes into the user application.
The user application can use any distro that has the tools compile for.
And it's not only compilation, its init script has to be compatible with all supported distributions.

tuler: This moved to PR https://github.com/sunodo/sunodo/pull/22

apps/cli/src/commands/build.ts
```diff
@@ -0,0 +1,296 @@
...
+    cmd: string[];
+    dataSize: string;
+    entrypoint: string[];
+    env: string[];
+    ramSize: string;
+    sdkVersion: string;
+    workdir: string;
+};
+
+const CARTESI_LABEL_PREFIX = "io.cartesi.rollups";
```
endersonmaia, line 23: I'm not sure if we need a machine namespace/hierarchy like `io.cartesi.rollups.machine`.

Imagine if we need to control other components from the cartesi-node that are related to the DApp image and need to be passed to, say, `dispatcher`, like the `network` (chain) or something else.

`io.cartesi.rollups.node.network` could hold this information.

I know we can change this in the future, won't block for it, but I think it deserved some thought/design.

tuler: I understand.

The network is a trickier example, that we will need to tackle soon. Because it will need to produce different machines.

apps/cli/src/sunodoCommand.ts
```diff
@@ -1,7 +1,26 @@
...
+export type Flags<T extends typeof Command> = Interfaces.InferredFlags<
+    (typeof SunodoCommand)["baseFlags"] & T["flags"]
+>;
+export type Args<T extends typeof Command> = Interfaces.InferredArgs<T["args"]>;
+
+export abstract class SunodoCommand<T extends typeof Command> extends Command {
+    protected flags!: Flags<T>;
+    protected args!: Args<T>;
+
+    protected buildStore: BuildStore = new BuildStore(process.cwd());
```
endersonmaia, line 13: Can you explain how is this being used?

tuler: I forgot to delete it. I was going to write a manager for the build artifacts, but then I decided to simplify and keep just one artifact.

endersonmaia:
```suggestion
```

apps/cli/src/utils/store.ts
```diff
@@ -0,0 +1,23 @@
...
+     * Deletes the contents of the store, but not the directory
+     */
+    public async clear(): Promise<void> {
+        await fs.emptyDir(this.location);
+    }
+
+    public pathFor(sub: string): string {
+        return path.join(this.location, sub);
+    }
+}
```
endersonmaia: Since it's going to be used, remove this.

```suggestion
```

endersonmaia (changes requested):
The first 2 commit should be removed from this PR.

Also, running the command I get errors:

```shell

```

apps/cli/src/commands/build.ts
```diff
@@ -0,0 +1,296 @@
...
+        const extraBytes = bytes.parse(info.dataSize);
+        const extraBlocks = Math.ceil(extraBytes / blockSize);
+        const extraSize = `+${extraBlocks}`;
+
+        await execa(
+            "docker",
+            [
+                "run",
+                "--rm",
+                "-v",
```
endersonmaia, line 186: Let's favor the long version for clarity.

```suggestion
                "--volume",
```

apps/cli/src/commands/build.ts
```diff
@@ -0,0 +1,296 @@
...
+        // join everything
+        const bootargs = [cwd, [env, entrypoint_cmd].join(" ")].join(";");
+        this.log(bootargs);
+
+        await execa(
+            "docker",
+            [
+                "run",
+                "--rm",
+                "-v",
```
endersonmaia, line 249:
```suggestion
                "--volume",
```

apps/cli/src/commands/shell.ts
```diff
@@ -0,0 +1,68 @@
...
+        const containerDir = "/mnt";
+        const bind = `${path.resolve(path.dirname(ext2Path))}:${containerDir}`;
+        const ext2 = path.join(containerDir, path.basename(ext2Path));
+        const ramSize = "128Mi";
+        const driveLabel = "root";
+        const toolchainImage = "sunodo/toolchain:0.14.0"; // XXX: how to resolve sdk version?
+        const args = [
+            "run",
+            "-it",
+            "-v",
```
endersonmaia, line 38:
```suggestion
            "--interactive",
            "--tty",
            "--volume",
```

endersonmaia (changes requested):
Made small suggestions on docker commands inside the code.

We should consider using `docker container run|create` instead of `docker run|create`, it's more explicit that we're interacting with containers and not with images. The same we do with `docker image inspect`.

We could also use `docker image build`.

See:

- https://docs.docker.com/engine/reference/commandline/container/
- https://docs.docker.com/engine/reference/commandline/image/

---

I'm still failing to execute the build, will send errors I'm getting on another comment.

endersonmaia:
I'm on linux and the `sunodo build` is creating a `.sunodo/image` directory with access only to the `root` user.

```
❯ ls -lha .sunodo
total 467M
drwxr-xr-x 3 endersonmaia endersonmaia 4,0K abr 25 20:53 .
drwxrwxr-x 4 endersonmaia endersonmaia 4,0K abr 25 20:53 ..
drwx------ 2 root         root         4,0K abr 25 20:53 image
-rw-r--r-- 1 root         root         246M abr 25 20:53 image.ext2
-rw------- 1 endersonmaia endersonmaia 223M abr 25 20:53 image.tar
```

endersonmaia (approved):

PR #22, opened 2023-04-20
Packaging of machine-emulator-tools as a docker image

tuler:
This packages a docker image from scratch with the contents of [machine-emulator-tools](https://github.com/cartesi/machine-emulator-tools) release.

The arch of the image is `linux/riscv64` and the intent is to be used by dapps Dockerfiles to add necessary files to the dapp, like so:

```
COPY --from sunodo/machine-emulator-tools / /
```

Eventually this could be published by Cartesi itself as part of their release process.

There is a fix to workaround a know bug of the location of the init script inside the package.

I already built and published the images to both DockerHub and GitHub

### DIFF
--- a/packages/machine-emulator-tools/Dockerfile
+++ b/packages/machine-emulator-tools/Dockerfile
@@ -6,4 +6,6 @@ ADD https://github.com/cartesi/machine-emulator-tools/releases/download/v${VERSI
 RUN mkdir -p /tmp/machine-emulator-tools && tar -C /tmp/machine-emulator-tools -xzf /tmp/machine-emulator-tools.tar.gz
 
 FROM scratch as tools
-COPY --from=download /tmp/machine-emulator-tools/ /
+COPY --from=download /tmp/machine-emulator-tools/sbin/init /usr/sbin/init
+COPY --from=download /tmp/machine-emulator-tools/etc/hostname /etc/hostname
+COPY --from=download /tmp/machine-emulator-tools/opt/cartesi /opt/cartesi/
--- a/packages/machine-emulator-tools/Dockerfile
+++ b/packages/machine-emulator-tools/Dockerfile
@@ -1,0 +1,9 @@
+# download machine-emulator tools and create a scratch image the tar.gz contents at root
+FROM busybox as download
+ARG VERSION
+ADD https://github.com/cartesi/machine-emulator-tools/releases/download/v${VERSION}/machine-emulator-tools-v${VERSION}.tar.gz \
+    /tmp/machine-emulator-tools.tar.gz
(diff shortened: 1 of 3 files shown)


packages/machine-emulator-tools/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# download machine-emulator tools and create a scratch image the tar.gz contents at root
+FROM busybox as download
+ARG VERSION
```
endersonmaia, line 3: If we don't use `bake`, I think we need a default value here too.

```suggestion
ARG VERSION=0.10.0
```

packages/machine-emulator-tools/README.md
```diff
@@ -0,0 +1,33 @@
...
+```dockerfile
+COPY --from=cartesi/machine-emulator-tools:0.10.0 / /
+```
+
+While `cartesi` doesn't provide such image, Sunodo will provide the image under the `ghcr.io/sunodo` organization.
+
+So the application should include the following:
+
+```dockerfile
+COPY --from=ghcr.io/sunodo/machine-emulator-tools:0.10.0-ubuntu22.04 / /
```
endersonmaia, line 18:
```suggestion
COPY --from=sunodo/machine-emulator-tools:0.10.0-ubuntu22.04 / /
```

packages/machine-emulator-tools/Dockerfile
tuler: Isn't it safer to not have a default value? Make it required

endersonmaia: I was thinking on the developer experience here.

But, at the same time, this isn't a thing the developer will interact, right?

I'll mark this comment as resolved.

packages/machine-emulator-tools/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+ADD https://github.com/cartesi/machine-emulator-tools/releases/download/v${VERSION}/machine-emulator-tools-v${VERSION}.tar.gz \
+    /opt/cartesi/share/machine-emulator-tools.tar.gz
+RUN tar -C /tmp -xzf /opt/cartesi/share/machine-emulator-tools.tar.gz
```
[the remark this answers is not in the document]
tuler: I want `/tmp` to have only the files I'll copy to the next stage (and not the .tar.gz)
Because in the future I'll turn the next stage to only `COPY --from=download /tmp /`
So I can't have the .tar.gz in there.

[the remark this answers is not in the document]
endersonmaia: So you coud untar it on a folder inside `/tmp/`, like `/tmp/machine-emulator-tools` and copy from there.

I know it's a temporary stage, but it feels weird to me copying code to `/opt/cartesi`.

endersonmaia:
@tuler LGTM

Please, squash so I can approve.

PR #25, opened 2023-04-25
Update genext2fs with better size calculation

tuler:
This updates the genext2fs with changes that include `libarchive` for correct size calculation when generating from `tar`.

In a future PR we should use a tagged version instead of the `cartesi` moving branch.
And there are a few other things that should be fixed in the `genext2fs` fork:

- `-r` behavior is strange, like `-r +0` doesn't work
- fix the `--version` to print a version of our fork
- update `--help` with the new args


endersonmaia (changes requested):
Need to remove the first commit.

endersonmaia (approved):

PR #40, opened 2023-04-29
Sunodo run command

tuler:
Very basic execution of the cartesi node.
Still needs a better feedback.
Or maybe the logs of the node could be much better.

What I'd like to see is something like:

```
Ethereum node ready: http://localhost:8545
GraphQL Server ready: http://localhost:8080/graphql
Inspect Server ready: http://localhost:8080/inspect
```


tuler:
This must use docker images built from https://github.com/cartesi-corp/rollups/tree/feature/localhost-rpc-url

apps/cli/src/node/devel/docker-compose.yml
```diff
@@ -0,0 +1,150 @@
...
+    POSTGRES_PASSWORD: postgres
+    POSTGRES_DB: postgres
+
+services:
+    anvil:
+        image: sunodo/anvil
+        environment:
+            ANVIL_IP_ADDR: 0.0.0.0
+        ports:
+            - "8545:8545"
```
endersonmaia, line 16:
```suggestion
            - "8545"
```

This way the host's port will be dynamic.

apps/cli/src/node/devel/docker-compose.yml
```diff
@@ -0,0 +1,150 @@
...
+        image: sunodo/cartesi-node:latest
+        depends_on:
+            set_interval_mining:
+                condition: service_completed_successfully
+        volumes:
+            - blockchain-data:/opt/cartesi/share/deployments:ro
+            - ./.sunodo/deployments:/deployments:ro
+            - ./.sunodo/image:/opt/cartesi/share/dapp-bin/0_0
+        ports:
+            - 8080:8080
```
endersonmaia, line 86:
```suggestion
            - "8080"
```

same here

endersonmaia (changes requested):
I'm having problems with `.sunodo/deployments` permissions, and I have to delete the folder everytime I need to do `sunodo run` again.

apps/cli/src/node/devel/docker-compose.yml
tuler: It's very incovenient to have a dynamic host port. This will typically be configured in the user metamask. It's a standard to be at http://localhost:8545

apps/cli/src/node/devel/docker-compose.yml
tuler: Again, also may be inconvenient. User will develop his frontend pointing to the graphql endpoint.

apps/cli/src/node/devel/docker-compose.yml
endersonmaia: We could print the port defined dinamcally, and also add an option to define the port from the `sunodo run` command.

tuler: I still think the default should be `8545`, not the dynamic one.

endersonmaia: A second `sunodo run` would result in an error. 

I prefer defaulting to a dynamic and free of conflict, and requiring the port number deifinition an explicit option.

tuler: Very unusual to run more than one application at the same time during development. It will be very confusing to not default to 8545 and the configuration of metamask.

endersonmaia:
@tuler can you do the same you did with `packages/anvil`?

Default the local build to `:devel` and add the build for `packages/toolchain` and `packages/machine-emulator-tools` to the CI?

apps/cli/src/commands/run.ts
```diff
@@ -0,0 +1,64 @@
...
+            "--project-name",
+            projectName,
+        ];
+
+        try {
+            // clean host compose environment, deployments host volume
+            await fs.emptyDir(path.join(".sunodo", "deployments"));
+
+            // run compose environment
+            await execa("docker", [...args, "up"], { stdio: "inherit" });
```
endersonmaia, line 56: Could we run `docker compose up --detach --wait` and run `docker compose logs --follow` later.

The idea is to remove the `anvil` noisy logs, and maybe add a `sunodo run run --verbose` argument to add the avil logs.

What do you think?

apps/cli/src/node/docker-compose.yml
```diff
@@ -0,0 +1,164 @@
...
+            test: ["CMD", "is_ready"]
+            interval: 10s
+            timeout: 1s
+            retries: 5
+        ports:
+            - 8080:8080
+        environment:
+            # global
+            RUST_LOG: info
+            # RUST_LOG: dispatcher=trace,info
```
endersonmaia, line 101:
```suggestion
```

apps/cli/src/node/docker-compose.yml
```diff
@@ -0,0 +1,164 @@
...
+            ## shared
+            ### redis
+            REDIS_ENDPOINT: redis://redis:6379
+            ### contract-address-file
+            DAPP_CONTRACT_ADDRESS_FILE: //opt/cartesi/share/deployments/dapp.json
```
endersonmaia, line 106:
```suggestion
            DAPP_CONTRACT_ADDRESS_FILE: /opt/cartesi/share/deployments/dapp.json
```

apps/cli/README.md
```diff
@@ -152,6 +153,28 @@ DESCRIPTION
...
+## `sunodo run`
+
+Run application node.
+
+```
+USAGE
+  $ sunodo run [--blockTime <value>]
+
+FLAGS
+  --blockTime=<value>  [default: 5] interval between blocks in seconds
```
endersonmaia, line 165: I miss the `--verbose` and `--epoch-duration` documentation here.

endersonmaia (changes requested):

apps/cli/README.md
tuler: This generated. I'll generate again.

endersonmaia:
@tuler LGTM, just waiting you to squash the fixups so I can approve.

tuler:
I have to fix the problem of copy the docker-compose.yml during build

endersonmaia (approved):

PR #45, opened 2023-05-12
Sunodo controller

tuler:
Implements a `sunodo controller` command.
The intention of the controller is to manage the execution of cartesi nodes for dapps by monitoring onchain activity of calls to dapp factories.

It follows the protocols defined by `IMachineProtocol` and `IFinancialProtocol` to keep records in a table of which dapps should be running and until when, and uses an API defined by `NodeDriver` to control the execution of nodes.

We will provide different implementations of `NodeDriver`, including one for local execution using Docker compose, and one for Kubernetes to be use in a cloud environment.


apps/cli/src/node/driver/k8s.ts
```diff
@@ -0,0 +1,64 @@
...
+        ]);
+
+        const minimumVersion = ">=3.0.0";
+        if (!semver.satisfies(helmVersion, minimumVersion)) {
+            throw new Error(`Helm version must be ${minimumVersion}`);
+        }
+    }
+
+    async start(dapp: DApp, location: string): Promise<NodeStatus> {
+        const name = dapp.address.substring(2, 8).toLocaleLowerCase();
```
endersonmaia, line 25:
```suggestion
        const name = dapp.address.substring(1, 7).toLocaleLowerCase();
```

a Service name can't start with a number

apps/cli/src/node/driver/k8s.ts
```diff
@@ -0,0 +1,64 @@
...
+        const chart = "oci://docker.io/cartesi/rollups-validator-node";
+        const httpProvider = "http://host.docker.internal:8545";
+        const wsProvider = "ws://host.docker.internal:8545";
+        const serverManager = "cartesi/server-manager:0.9.1";
```
endersonmaia, line 29:
```suggestion
        const serverManager = "cartesi/server-manager:0.7.0";
```

endersonmaia: this won't work, since the current rollups-validator-node helm charts requires a DApp image that's based on the server-manager and embeds the machine-snapshot.

I shall work on a fixup for that to be optional.

apps/cli/src/node/driver/k8s.ts
```diff
@@ -0,0 +1,64 @@
...
+        };
+
+        // TODO: serverManager.initContainers for IPFS download
+        // TODO: serverManager.extraVolumes for cartesi machine
+
+        const set = Object.entries(values)
+            .map(([k, v]) => `${k}=${v}`)
+            .join(",");
+        const flags = ["--set", set];
+        const cmd = await execa("helm", ["install", name, chart, ...flags], {
```
endersonmaia, line 52:
```suggestion
        const cmd = await execa("helm", ["upgrade", "--install",name, chart, ...flags], {
```

This way, helm will upgrade existing charts and install them if a chart with that name doesn't exist.

endersonmaia (changes requested):

tuler:
Opted to implement a k8s driver using the CRD approach.
Depend on the `cartesi-rollups-operator`, which will act upon the custom resource.

endersonmaia (approved):
The code looks ok, I still need to put it to run.

endersonmaia:
@tuler can you write the instructions on how to run and test this PR?

I can run the controller, but I get some errors, will need your help to run it locally.

endersonmaia (changes requested):
What happens if the Application CR is deleted (by hand) from the cluster, should the controller install it again if there are still funds?

Or is this the operator responsibility? :thinking:

apps/cli/test/node/database.test.ts
```diff
@@ -0,0 +1,194 @@
...
+    test("store and load at future time", () => {
+        const db = new Database();
+        const now: bigint = BigInt(Date.now());
+        db.addMachine(1n, "0xdeadbeef", "C123");
+        db.addMachine(1n, "0x123", "Que2");
+        db.addApplication(now, {
+            blockNumber: 2n,
+            blockHash: "0xdeadbeef",
+            transactionHash: "0xdeadbeef",
+            address: "0xdeadbeef",
```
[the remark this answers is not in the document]
tuler: I don't think we need to check that. These values come from the blockchain, and they will be "valid"

tuler:
If the CR is deleted by hand the controller will not recreate it.
The controller saves in his database the blockNumber he is at.
If you delete this database, and scan from the beginning it will recreate.

I don't think this is something we have to tackle now.

endersonmaia:
> I don't think this is something we have to tackle now.

But we should tackle this in the future. 

We may need/want to pass status (or any other kind of) information  from the operator to the CR so that the controller act upon it.

endersonmaia (approved):
LGTM

endersonmaia (approved):
Still getting errors from remote-cartesi-machine, but will approve it since it's not in the scope of this PR.

endersonmaia:
> Still getting errors from remote-cartesi-machine, but will approve it since it's not in the scope of this PR.

Nevermind, I was using the wrong versions for SDK  and emulator.

PR #49, opened 2023-05-18
CLI Debian publishing

omidasadpour:
Here's a step-by-step breakdown of the provided Terraform code:

**Step 1:** S3 Bucket Configuration

- Creates an S3 bucket named `cli-prod-sunodo-origin` with the specified name and tags.
- Sets the bucket ACL to "private" using `aws_s3_bucket_acl` resource.
- Configures bucket ownership controls to allow ObjectWriter access using `aws_s3_bucket_ownership_controls` resource.
- Sets up public access block configuration to disallow public access to the bucket using `aws_s3_bucket_public_access_block` resource.

**Step 2:** IAM Configuration

- Defines an IAM policy document using `data.aws_iam_policy_document` to specify the required actions and resources.
- Creates an IAM policy named `sunodo_cli` with the defined policy document using `aws_iam_policy` resource.
- Creates an IAM identity_providers for github OIDC

**Step 3:** CloudFront Configuration

- Creates a CloudFront origin access identity named `sunodo_cli_origin_access_identity` using `aws_cloudfront_origin_access_identity` resource.
- Defines an IAM role named `cloudfront-iam-role` with an assume role policy allowing CloudFront service to assume the role using `aws_iam_role` resource.
- Defines an IAM policy named `cloudfront-iam-policy` with the required permissions using `aws_iam_policy` resource.
- Attaches the IAM policy to the IAM role using `aws_iam_role_policy_attachment` resource.
- Creates an ACM certificate for the domain `cli.sunodo.io` and `*.sunodo.io` using `aws_acm_certificate` resource. (**wildcard**)
- Creates a CloudFront distribution named `sunodo_cli`_distribution with the specified settings, including the S3 bucket origin, viewer certificate, cache behavior, and restrictions using `aws_cloudfront_distribution` resource.
- Associates the CloudFront distribution with the subdomain `cli.sunodo.io` using a **CNAME** record in Route 53 using `aws_route53_record` resource.

### DIFF
--- a/.github/workflows/cli.yaml
+++ b/.github/workflows/cli.yaml
@@ -23,31 +23,19 @@ concurrency:
 permissions:
     contents: read
     packages: write
+    id-token: write
+env:
+  TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+  TURBO_TEAM: ${{ vars.TURBO_TEAM }}
 jobs:
     build:
         runs-on: ubuntu-latest
-        env:
-            TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
-            TURBO_TEAM: ${{ vars.TURBO_TEAM }}
         steps:
             - name: Checkout
               uses: actions/checkout@v4
 
-            - name: Setup Node.js 20
-              uses: actions/setup-node@v3
-              with:
-                  node-version: 20
-                  registry-url: "https://registry.npmjs.org"
-                  cache: "yarn"
-
-            - name: Install Foundry
-              uses: foundry-rs/foundry-toolchain@v1
-
-            - name: Install dependencies
-              run: yarn install
-
             - name: Build
-              run: yarn build --filter @sunodo/cli
+              uses: './.github/actions/cli-ci-setup'
 
             - name: Publish
               if: ${{ inputs.release }}
@@ -55,3 +43,112 @@ jobs:
               run: npm publish
               env:
                   NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+    pack_upload_deb:
+          runs-on: ubuntu-latest
+          if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
+          steps:
+            - uses: actions/checkout@v3
+
+            - name: Setup environment
+              uses: './.github/actions/cli-ci-setup'
+
+            - name: Configure AWS Credentials
+              uses: aws-actions/configure-aws-credentials@v2
+              with:
+                  role-to-assume: ${{ secrets.IAM_ROLE_ARN }}
+                  aws-region: ${{ secrets.AWS_REGION }}
+
+            - name: Install GPG key for debian package signing
+              id: gpg_key
+              uses: crazy-max/ghaction-import-gpg@v5
+              with:
+                  gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+                  passphrase: ${{ secrets.GPG_PASSPHRASE }}
+
+            - name: Build debian package
+              run: yarn oclif pack deb --root="./apps/cli"
+              env:
+                  SUNODO_DEB_KEY: ${{ steps.gpg_key.outputs.keyid }}
+
+            - name: Upload debian packages to Github artifacts
+              uses: actions/upload-artifact@v3
+              with:
+                  name: Packages-deb
+                  path: ./apps/cli/dist/deb/*.deb
+
+            - name: Upload debian packages to S3
+              working-directory: apps/cli
+              run: |
+                yarn oclif upload deb
+
+    pack_upload_tarballs:
+      runs-on: ubuntu-latest
+      if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
+      steps:
+            - uses: actions/checkout@v3
+
+            - name: Setup environment
+              uses: './.github/actions/cli-ci-setup'
+
+            - name: Configure AWS Credentials
+              uses: aws-actions/configure-aws-credentials@v2
+              with:
+                  role-to-assume: ${{ secrets.IAM_ROLE_ARN }}
+                  aws-region: ${{ secrets.AWS_REGION }}
+
+            - name: Building tarballs
+              run: yarn oclif pack tarballs --root="./apps/cli"
+
+            - name: Upload tarball packages to Github artifacts
+              uses: actions/upload-artifact@v3
+              with:
+                  name: Packages-tarballs
+                  path: ./apps/cli/dist/*.tar.*
+
+            - name: Upload tarball packages to S3
+              working-directory: apps/cli
+              run: |
+                yarn oclif upload tarballs
+
+    promote-deb-and-tarballs:
+      needs: [pack_upload_deb, pack_upload_tarballs]
+      runs-on: ubuntu-latest
+      steps:
+            - uses: actions/checkout@v3
+
+            - name: Setup environment
+              uses: './.github/actions/cli-ci-setup'
+
+            - name: Configure AWS Credentials
+              uses: aws-actions/configure-aws-credentials@v2
+              with:
+                  role-to-assume: ${{ secrets.IAM_ROLE_ARN }}
+                  aws-region: ${{ secrets.AWS_REGION }}
+
+            - name: Download deb packages from Github artifacts
+              uses: actions/download-artifact@v3
+              with:
+                name: Packages-deb
+                path: apps/cli/dist
+
+            - name: List all the downloaded files (for debugging)
+              run: ls -R
+              working-directory: apps/cli/dist
+
+            - name: Extract version and SHA
+              id: extract_info
+              run: |
+                filename=$(find . -name '*amd64.deb' -type f)
+                version=$(echo "$filename" | sed -n 's/.*_\([0-9]*\.[0-9]*\.[0-9]*\).*-.*\.deb/\1/p')
+                sha=$(echo "$filename" | sed -n 's/.*\.\([^-]*\)-.*\.deb/\1/p')
+
+                echo "::set-output name=version::$version"
+                echo "::set-output name=sha::$sha"
+              working-directory: apps/cli/dist
+
+            - name: Promote packages to S3
+              run: |
+                yarn oclif promote --deb --xz --indexes --version=${{ steps.extract_info.outputs.version }} --sha=${{ steps.extract_info.outputs.sha }}
+                aws s3 cp  s3://cli-prod-sunodo-origin/versions/${{ steps.extract_info.outputs.version }}/${{ steps.extract_info.outputs.sha }}/apt/./ s3://cli-prod-sunodo-origin/channels/stable/apt/./ --recursive
+              working-directory: apps/cli
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -83,6 +83,13 @@
             "@oclif/plugin-update"
         ],
         "topicSeparator": " ",
+        "update": {
+            "s3": {
+                "xz": true,
+                "bucket": "cli-prod-sunodo-origin",
+                "host": "https://cli.sunodo.io"
+            }
+        },
         "macos": {
             "identifier": "io.sunodo.cli"
         }
--- a/apps/cli/tf/bucket.tf
+++ b/apps/cli/tf/bucket.tf
@@ -1,0 +1,82 @@
+module "acm_request_certificate" {
+  source = "cloudposse/acm-request-certificate/aws"
+
+  # https://github.com/cloudposse/terraform-aws-acm-request-certificate
+  version                           = "v0.17.0"
+  domain_name                       = "cli.sunodo.io"
+  process_domain_validation_options = true
+  ttl                               = "300"
+  zone_id                           = var.sunodo_prod_route53_zone_id
+}
+
+module "cdn" {
+  source = "cloudposse/cloudfront-s3-cdn/aws"
+
+  # https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/tree/0.90.0
+  version                             = "v0.90.0"
+  name                                = "sunodo"
+  namespace                           = "cli"
+  stage                               = "prod"
+  aliases                             = ["cli.sunodo.io"]
+  dns_alias_enabled                   = true
+  parent_zone_id                      = var.sunodo_prod_route53_zone_id
+  acm_certificate_arn                 = module.acm_request_certificate.arn
+  cloudfront_access_log_create_bucket = false
+  cloudfront_access_logging_enabled   = false
+  allowed_methods                     = ["GET", "HEAD", "OPTIONS"]
+  cached_methods                      = ["GET", "HEAD"]
+  versioning_enabled                  = false
+  block_origin_public_access_enabled  = false
+
+  depends_on = [module.acm_request_certificate]
+}
+
+data "aws_iam_policy_document" "sunodo_cli_s3_doc" {
+  statement {
+    actions = [
+      "s3:ListBucket",
+    ]
+    resources = [
+      "arn:aws:s3:::cli-prod-sunodo-origin",
+    ]
+  }
+  statement {
+    actions = [
+      "s3:PutObject",
+      "s3:GetObject",
+      "s3:DeleteObject",
+      "s3:PutBucketPolicy",
+      "s3:GetBucketPolicy",
+      "s3:PutBucketAcl",
+      "s3:GetBucketAcl",
+      "s3:PutObjectAcl",
+      "s3:ListMultipartUploadParts",
+      "s3:AbortMultipartUpload",
+      "s3:GetObjectAcl",
+      "s3:GetObjectTagging",
+      "s3:PutObjectTagging",
+    ]
+    resources = [
+      "arn:aws:s3:::cli-prod-sunodo-origin/*",
+    ]
+  }
+}
+
+resource "aws_iam_policy" "sunodo_cli_s3_policy" {
+  name        = "sunodo_cli_s3_policy"
+  path        = "/"
+  description = "Sunodo CLI S3 Storage Policy"
+  policy      = data.aws_iam_policy_document.sunodo_cli_s3_doc.json
+}
+
+module "iam_iam-assumable-role-with-oidc" {
+  # https://github.com/terraform-aws-modules/terraform-aws-iam/tree/v5.20.0/modules/iam-assumable-role-with-oidc
+  source                         = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version                        = "5.20.0"
+  create_role                    = true
+  role_name                      = "sunodo_cli"
+  provider_url                   = "token.actions.githubusercontent.com"
+  oidc_fully_qualified_audiences = ["sts.amazonaws.com"]
+  oidc_subjects_with_wildcards   = ["repo:sunodo/sunodo:*"]
+  role_policy_arns               = [aws_iam_policy.sunodo_cli_s3_policy.arn]
+}
--- a/apps/cli/tf/outputs.tf
+++ b/apps/cli/tf/outputs.tf
@@ -1,0 +1,4 @@
+output "GITHUB_IAM_ROLE_ARN" {
+  value       = module.iam_iam-assumable-role-with-oidc.iam_role_arn
+  description = "Connect to the S3 bucket from github with this arn"
+}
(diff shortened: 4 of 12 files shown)


apps/cli/tf/bucket.tf
```diff
@@ -0,0 +1,108 @@
...
+
+  tags = {
+    Name        = "sunodo_cli_bucket"
+    Environment = "Production"
+  }
+}
+
+resource "aws_s3_bucket_acl" "sunodo_cli_bucket_acl" {
+  bucket = aws_s3_bucket.sunodo_cli_bucket.id
+  acl    = "public-read"
```
tuler, line 13: I don't think the bucket itself must be public read.
It's better to have some domain (`cli.sunodo.io`) in a CloudFront web distribution, and S3 behind it.

oap75: Agree, working on it ...

oap75: Fixed !

tuler:
Looks good. I suggest we build on top of this PR before merging to main.

I still think tf code is so verbose, compared to [cdk](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudfront-readme.html#from-an-s3-bucket).

endersonmaia:
> I still think terraform code is so verbose, compared to [cdk](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudfront-readme.html#from-an-s3-bucket).

Yes, it's more aligned with the AWS API, like CloudFormation, CDK has a lot of abstractions on top of the CloudFormation.

tuler:
> Yes, it's more aligned with the AWS API, like CloudFormation, CDK has a lot of abstractions on top of the CloudFormation.

But if terraform had good composability it was about time to have layers of abstraction as well to make a s3+cloudfront distribution with 10 lines of code

oap75:
> It's possible to create/reuse [terraform modules](https://developer.hashicorp.com/terraform/language/modules) and there are public registries like this https://registry.terraform.io/namespaces/terraform-aws-modules.

Yeah, We can even create our own terraform modules .

tuler:
> > It's possible to create/reuse [terraform modules](https://developer.hashicorp.com/terraform/language/modules) and there are public registries like this https://registry.terraform.io/namespaces/terraform-aws-modules.
> 
> Yeah, We can even create our own terraform modules .

Ok, but the idea is to write as less code as possible, and reuse. Is it possible to write minimal cloudfront distribution + s3 code using [cloudfront module](https://registry.terraform.io/modules/terraform-aws-modules/cloudfront/aws/latest)?

endersonmaia:
I found this module https://registry.terraform.io/modules/cloudposse/cloudfront-s3-cdn/aws/latest

I don't know the reputation of the publishers but looks like it has what you want.

oap75:
> the idea is to write as less code as possible, and reuse. Is it possible to write minimal cloudfront distribution + s3 code using [cloudfront module](https://registry.terraform.io/modules/terraform-aws-modules/cloudfront/aws/latest)?

Yes, it is possible to reduce code duplication and reuse existing modules to create a minimal CloudFront distribution with an S3 bucket. By using a Terraform module specifically designed for CloudFront and S3 integration, we can abstract away the complexity and write less code.

oap75:
> I don't know the reputation of the publishers

That's why I suggested to create our own modules. we can even publish them as an opensource modules too.

oap75:
> > > It's possible to create/reuse [terraform modules](https://developer.hashicorp.com/terraform/language/modules) and there are public registries like this https://registry.terraform.io/namespaces/terraform-aws-modules.
> > 
> > 
> > Yeah, We can even create our own terraform modules .
> 
> Ok, but the idea is to write as less code as possible, and reuse. Is it possible to write minimal cloudfront distribution + s3 code using [cloudfront module](https://registry.terraform.io/modules/terraform-aws-modules/cloudfront/aws/latest)?

Fixed ! 

1. [cloudposse/acm-request-certificate/aws](https://github.com/cloudposse/terraform-aws-acm-request-certificate) module to create AWS ACM for cloudfront, 

2. [cloudposse/cloudfront-s3-cdn/aws](https://github.com/cloudposse/terraform-aws-cloudfront-s3-cdn/tree/0.90.0) module to create S3 bucket + cloudformation 

3. [cloudposse/iam-system-user/aws](https://github.com/cloudposse/terraform-aws-iam-system-user) module to create oclif IAM user to upload packages to S3

tuler:
LGTM
Add a changeset by running `yarn changeset` and selecting "minor" change to @sunodo/cli with change "add infrastructure code for CLI tarballs and debian release`

oap75:
> LGTM Add a changeset by running `yarn changeset` and selecting "minor" change to @sunodo/cli with change "add infrastructure code for CLI tarballs and debian release`

Okay 
I think we don't need the [oclif_user](https://github.com/sunodo/sunodo/pull/49/commits/00d261a830486ba3f4789228c30a0e4d7a359baa) module anymore ! Am I right ?

apps/cli/tf/bucket.tf
```diff
@@ -0,0 +1,67 @@
...
+  cloudfront_access_logging_enabled   = false
+  allowed_methods                     = ["GET", "HEAD", "OPTIONS"]
+  cached_methods                      = ["GET", "HEAD"]
+  versioning_enabled                  = false
+  block_origin_public_access_enabled  = false
+
+  depends_on = [module.acm_request_certificate]
+}
+
+module "oclif_user" {
```
tuler, line 35: We don't need a user

tuler (commented):
Create a IAM role with write permission to the bucket.

Then we can use AssumeRole with the OIDC provider and [configure-aws-credentials](https://github.com/aws-actions/configure-aws-credentials#usage)

apps/cli/tf/bucket.tf
```diff
@@ -0,0 +1,73 @@
+module "acm_request_certificate" {
+  source = "cloudposse/acm-request-certificate/aws"
+
+  # https://github.com/cloudposse/terraform-aws-acm-request-certificate
+  version                           = "v0.17.0"
+  domain_name                       = "cli.sunodo.io"
+  process_domain_validation_options = true
+  ttl                               = "300"
+  zone_id                           = "Z09327832YAJGJ5APMP08"
```
tuler, line 9: Is the `zone_id` sensitive information to be here?

apps/cli/tf/bucket.tf
```diff
@@ -0,0 +1,73 @@
...
+}
+
+resource "aws_iam_policy" "sunodo_cli_s3_policy" {
+  name        = "sunodo_cli_s3_policy"
+  path        = "/"
+  description = "Sunodo CLI S3 Storage Policy"
+  policy      = data.aws_iam_policy_document.sunodo_cli_s3_doc.json
+}
+
+module "github-oidc" {
```
tuler, line 61: If the OIDC is created here what would happen if other future resources need it?
Maybe the terraform code should be in `packages/terraform` and not under `apps/cli/tf`.
Or some module organization terraform provides.

apps/cli/tf/bucket.tf
oap75: I think yes, 

It is generally recommended to treat the zone_id as sensitive information and avoid sharing it in public forums, code repositories, or any other insecure channels.

I'll fix it ...

apps/cli/tf/bucket.tf
oap75: The module gets an ENV related to the `repositories`. I think we can have multiple OIDCs per each repo . 
But, your right if we just want to use one OIDC in multiple resources in the same repo.

tuler:
I create a terraform cloud account and invited you both.
Still need to configure the trust between AWS and Terraform Cloud.
https://developer.hashicorp.com/terraform/cloud-docs/workspaces/dynamic-provider-credentials/aws-configuration

oap75:
> I create a terraform cloud account and invited you both. Still need to configure the trust between AWS and Terraform Cloud. https://developer.hashicorp.com/terraform/cloud-docs/workspaces/dynamic-provider-credentials/aws-configuration

One solution is to define variables and marks them as `Sensitive` in TF cloud

oap75:
> The apply role has write permissions to S3 and Cloudfront.

It also needs permission to create IAM policy and IAM role, Create Cname and A record in route53 and Create ACM certificate

apps/cli/tf/bucket.tf
```diff
@@ -0,0 +1,71 @@
...
+  # https://github.com/terraform-module/terraform-aws-github-oidc-provider
+  source                    = "terraform-module/github-oidc-provider/aws"
+  version                   = "~> 2.1.0"
+  oidc_provider_arn         = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:oidc-provider/token.actions.githubusercontent.com"
```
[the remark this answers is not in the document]
oap75: Fixed by using another module !

apps/cli/tf/bucket.tf
```diff
@@ -0,0 +1,70 @@
...
+    actions = [
+      "s3:ListBucket",
+    ]
+    resources = [
+      "arn:aws:s3:::cli-prod-sunodo-origin",
+    ]
+  }
+  statement {
+    actions = [
+      "s3:*",
```
endersonmaia, line 45: Can this be more restrict/specific?

apps/cli/tf/outputs.tf
```diff
@@ -0,0 +1,4 @@
+output "GITHUB_IAM_ROLE_ARN" {
+value = module.iam_iam-assumable-role-with-oidc.iam_role_arn
+description = "Connect to the S3 bucket from github with this arn"
```
endersonmaia, line 3: Proper indentation.

```suggestion
  value = module.iam_iam-assumable-role-with-oidc.iam_role_arn
  description = "Connect to the S3 bucket from github with this arn"
```

Let's consider using a linter/formatter, if tf doesn't have one, we could use the current prettier configuration in the repository if it supports tf.

endersonmaia (changes requested):
I suggest you to create a `./apps/cli/tf/README.md` file explaining the requirements needed for this to work.

- like Terraform Cloud account
- required tf variables
- GitHub environment/secrets used

apps/cli/tf/variables.tf
```diff
@@ -0,0 +1,11 @@
+variable "sunodo_cli_region" {
+  description = "The sunodo cli AWS region"
+  type        = string
+  sensitive   = true
+}
+
+variable "sunodo_cli_zone_id" {
+  description = "The sunodo cli zone ID"
```
endersonmaia, line 8: Maybe use `sunodo_cli_dns_zone_id` or `sunodo_cli_route53_zone_id`. I'm afraid zone ID is too generic and could be used as names for other resources in AWS. 

What do you think?

endersonmaia: Ok, I'm assuming the zone ID here points to a `sunodo.io` zone, and `cli.sunodo.io` is just a CNAME entry pointing to CloudFront's FQDN.

Agreed with removing the `_cli_`  part.

apps/cli/tf/outputs.tf
oap75: We have these options : 

1 - **Terraform fmt**: 
`Terraform includes its built-in formatter called terraform fmt. It automatically updates and formats our Terraform code according to the Terraform style conventions. we can run terraform fmt command in our project directory to format the code.`

2 - **TFLint**: 
`TFLint is a Terraform linter that analyzes our code and provides suggestions, warnings, and errors based on a set of predefined rules. It helps identify potential issues or deviations from best practices. we can integrate TFLint into our CI/CD pipeline or run it locally. More information about TFLint can be found at: https://github.com/terraform-linters/tflint`

3 - **Checkov**: 
`Checkov is an open-source static analysis tool for infrastructure-as-code (IaC) files. It supports multiple IaC frameworks, including Terraform, and scans our code to identify security and compliance issues. Checkov can be integrated into our development workflow or run as a standalone tool. More information about Checkov can be found at: https://www.checkov.io/`

4 - **pre-commit**: 
`pre-commit is a framework that allows us to automatically run various linters, including Terraform linters, before each commit. It helps enforce code quality standards and prevents committing code that doesn't pass the linter checks. we can configure pre-commit to run Terraform linters like terraform fmt, TFLint, or Checkov. More information about pre-commit can be found at: https://pre-commit.com/`

apps/cli/tf/bucket.tf
oap75: I couldn't find any related docs in the oclif document. but, I'll check some restrictions.

tuler:
Both `amd64` and `arm64` were built and uploaded as build artifacts, but for some reason only the `amd64` package was uploaded by `oclif upload deb`.

Install of debian package in my machine failed, probably because there is no `arm64` package.

.github/workflows/cli.yaml
```diff
@@ -41,3 +42,178 @@ jobs:
...
+                sha=$(echo "$filename" | sed -n 's/.*\.\([^-]*\)-.*\.deb/\1/p')
+
+                echo "::set-output name=version::$version"
+                echo "::set-output name=sha::$sha"
+              working-directory: apps/cli/dist
+
+            - name: Promote packages to S3
+              run: |
+                yarn oclif promote --deb --xz --indexes --version=${{ steps.extract_info.outputs.version }} --sha=${{ steps.extract_info.outputs.sha }}
+                aws s3 cp  s3://cli-prod-sunodo-origin/versions/${{ steps.extract_info.outputs.version }}/${{ steps.extract_info.outputs.sha }}/apt/./ s3://cli-prod-sunodo-origin/channels/stable/apt/./ --recursive
```
tuler, line 218: do we need the extra `./` in those URLs?

apps/cli/package.json
```diff
@@ -83,7 +82,7 @@
         "eslint-config-oclif": "^4",
         "eslint-config-oclif-typescript": "^1",
         "npm-run-all": "^4",
-        "oclif": "^3.9",
+        "oclif": "^3.11.1",
```
tuler, line 85: Use `^3.11` instead of `^3.11.1`

apps/cli/scripts/install-debial.sh
```diff
@@ -0,0 +1,40 @@
...
+  dpkg -s apt-transport-https 1>/dev/null 2>/dev/null || \
+    (echo "" > /etc/apt/sources.list.d/sunodo.list \
+      && apt-get update \
+      && apt-get install -y apt-transport-https gnupg2 curl)
+
+  # add sunodo repository to apt
+  echo "deb https://cli.sunodo.io/channels/stable/apt ./" > /etc/apt/sources.list.d/sunodo.list
+
+  # install sunodo's release key for package verification
+  curl https://cli.sunodo.io/release.key | apt-key add -
```
tuler, line 27: this method of adding the key is deprecated.
Check this blog post
https://itsfoss.com/apt-key-deprecated/

.github/workflows/cli.yaml
```diff
@@ -41,3 +42,178 @@ jobs:
 
             - name: Build
               run: yarn build --filter @sunodo/cli
+
+    pack_upload_deb:
```
tuler, line 46: Do we want to run these new jobs also on PR, or just on tag?
The workflow also runs on PRs.

Also I see there are 3 jobs, and quite some code duplication. Maybe we could all in one job?

.github/workflows/cli.yaml
oap75: Yes, we need to add `./` contents into `./` of the stable channel.

apps/cli/package.json
```diff
@@ -63,7 +63,6 @@
         "@graphql-codegen/typescript-operations": "^4",
         "@graphql-typed-document-node/core": "^3.2",
         "@oclif/test": "^2.3",
-        "@sunodo/contracts": "*",
```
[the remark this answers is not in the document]
oap75: Yes, that's why I created a DELETE commit to test the upload process. 
I need your help to solve the problem.

.github/workflows/cli.yaml
oap75: Just on tag I think. 
That's why I added `if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')` section.

oap75: For the duplications. I was thinking to use job template.

.github/workflows/cli.yaml
endersonmaia, line 1: I think this JOB should only run when we have a tag.

So the only trigger here should be `on.push.tags["@sunodo/cli@*"]`

.github/workflows/cli.yaml
endersonmaia: I think we need a single job that runs only for the `on.push.tags` trigger.

oap75:
```
on:
  workflow_call:

```

Then use it in each jobs steps : 

```    uses:.//.github/workflows/common-cli.yml```

**Review by endersonmaia — CHANGES_REQUESTED** [review]
I'd like to see the terraform content on a separate commit.

Also the Github Actions code should be done in a single Job, I think.

I'd postpone publishing a script that makes the installation via `curl ...install.sh | bash`, for security/portability reasons.

We could add information in the docs on how to add the repository and install using apt-get by hand.

**Comment by oap75** [comment]
> I'd like to see the terraform content on a separate commit.

It was separated in the first place, then @tuler asked me to put them all in one commit.

File: .github/workflows/cli.yaml (hunk 12 above)
Comment (endersonmaia) [reply .github/workflows/cli.yaml:46]: It makes no sense to have each of these steps as independent Jobs since they won't need to run alone.

They are just steps that run every time, right?

**Comment by endersonmaia** [comment]
> > I'd like to see the terraform content on a separate commit.
> 
> It was separated in the first place, then @tuler asked me to put them all in one commit.

Yeah, we don't agree all the time.

Do what you think is best.

**Comment by tuler** [comment]
> > > I'd like to see the terraform content on a separate commit.
> > It was separated in the first place, then @tuler asked me to put them all in one commit.

Same PR, not necessarily same commit.

**Review by endersonmaia — CHANGES_REQUESTED** [review]

**Comment by oap75** [comment]
> We'd have to measure to be sure about it.

Tarball takes 42 minutes and debian takes about 15 minutes. So, if we add them in a row, it will take about 60 minutes. but in parallel it will take 42 minutes.

File: .github/workflows/cli.yaml [hunk 14]
```diff
@@ -107,34 +72,11 @@ jobs:
     pack_upload_tarballs:
       runs-on: ubuntu-latest
       if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
-      env:
-        TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
-        TURBO_TEAM: ${{ vars.TURBO_TEAM }}
       steps:
             - uses: actions/checkout@v3
```
endersonmaia, line 76: Do we need this if this is done at the composite action?

.github/workflows/cli.yaml
```diff
@@ -162,28 +104,8 @@ jobs:
...
```
endersonmaia, line 105: Do we need this if this is done at the composite action?

.github/workflows/cli.yaml
oap75: Yes, we need checkout to get the contents of the repo and then refer to the composition action

oap75:
> One last thing, shouldn't the `pack_*` jobs depend on the `build` job?
> 
> image

The strategy is to run build on each PR but run the rest og the jobs whenever we have a new tag `if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')`

.github/workflows/cli.yaml
endersonmaia: Indeed. What about the checkout inside the composite, is that needed?

.github/workflows/cli.yaml
oap75: Yes, we need checkout to get the contents of the repo and then refer to the composition action

endersonmaia:
> whenever

yeah, but it doesn't make send to `pack_*` if the `build` job fails, right?

that's why I insist in making it a dependency

oap75:
> yeah, but it doesn't make send to `pack_*` if the `build` job fails, right?
> 
> that's why I insist in making it a dependency

Inside the `pack_* jobs` we should use `composite` action as like as the `build` job. so, if we make them depandent on the `build` job, we will increase our CI time. (the `build` should run then the `pack_* `jobs should run)

Your right If the `build` job runs steps that are totally unique to the `pack_*` jobs.

endersonmaia:
@oap75 this branch needs to be rebased

tuler:
rebased

tuler:
I think this is too complex. We will reconsider distribution strategies in the future.

PR #51, opened 2023-05-22
Contracts deployment using anvil_loadState

tuler:
This provides tools for dumping and loading anvil state.

Instead of deploying contracts using `cartesi/rollups-hardhat:0.9.0` during node startup, which is a `hardhat` project deployment, we could do that during build time, dump the state of anvil to a file, and distribute the file as part of sunodo CLI.

The during node startup we can just load the state from the file to anvil.

The files generated by `hardhat`, `localhost.json` and the `deployments/localhost` directory could be also packaged during build time.

### DIFF
--- a/.changeset/popular-needles-beam.md
+++ b/.changeset/popular-needles-beam.md
@@ -1,0 +1,7 @@
+---
+"anvil": major
+---
+
+-   Changing Docker image from ENTRYPOINT to CMD (breaking change)
+-   Adding `eth_dump` and `eth_load` tools
+-   Change CI to use monorepo tags like `anvil@<version>`
--- a/packages/anvil/eth_dump
+++ b/packages/anvil/eth_dump
@@ -1,0 +1,11 @@
+#!/usr/bin/env sh
+
+# dump the state from the JSON-RPC server, convert from the hex anvil returns to text, which is a JSON string
+curl \
+    -sL \
+    -X POST \
+    -H 'Content-Type: application/json' \
+    --data '{"id":1,"jsonrpc":"2.0","method":"anvil_dumpState","params":[]}' "${RPC_URL:-http://127.0.0.1:8545}" | \
+jq -r .result| \
+cut -c 3- | \
+xxd -r -p
--- a/packages/anvil/eth_load
+++ b/packages/anvil/eth_load
@@ -1,0 +1,14 @@
+#!/usr/bin/env sh
+set -e
+RPC_URL="${RPC_URL:-http://127.0.0.1:8545}"
+
+# use a large column value so we have everything on one line, even though man says 256 is the maximum
+STATE=$(xxd -ps -c 1000000000)
+
+# build the JSON-RPC request, and write to a file because it's too long for the curl call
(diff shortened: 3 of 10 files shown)


endersonmaia:
@tuler shouldn't this PR point to another branch?

tuler:
I brought the changes to be base on `main`

packages/anvil/package.json
```diff
@@ -0,0 +1,5 @@
+{
+    "name": "anvil",
+    "private": true,
+    "version": "1.0.0"
```
endersonmaia, line 4: 1.0.0 already?

packages/anvil/eth_load
```diff
@@ -0,0 +1,14 @@
...
+# use a large column value so we have everything on one line, even though man says 256 is the maximum
+STATE=$(xxd -ps -c 1000000000)
+
+# build the JSON-RPC request, and write to a file because it's too long for the curl call
+DATA="{\"id\":2,\"jsonrpc\":\"2.0\",\"method\":\"anvil_loadState\",\"params\":[\"0x$STATE\"]}"
+TMPFILE=$(mktemp)
+echo $DATA > $TMPFILE
+
+echo "Loading state into ${RPC}"
+curl -sL -H 'Content-Type: application/json' -d @$TMPFILE $RPC | jq -r .result
```
endersonmaia, line 14:
```suggestion
#!/usr/bin/env sh
set -e
RPC_URL="${RPC_URL:-http://127.0.0.1:8545}"

# use a large column value so we have everything on one line, even though man says 256 is the maximum
STATE=$(xxd -ps -c 1000000000)

# build the JSON-RPC request, and write to a file because it's too long for the curl call
DATA="{\"id\":2,\"jsonrpc\":\"2.0\",\"method\":\"anvil_loadState\",\"params\":[\"0x$STATE\"]}"
TMPFILE=$(mktemp)
echo "$DATA" > "$TMPFILE"

echo "Loading state into ${RPC}"
curl -sL -H 'Content-Type: application/json' -d @"$TMPFILE" "$RPC_URL" | jq -r .result
```

packages/anvil/eth_dump
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env sh
+
+# dump the state from the JSON-RPC server, convert from the hex anvil returns to text, which is a JSON string
+curl \
+    -sL \
+    -X POST \
+    -H 'Content-Type: application/json' \
+    --data '{"id":1,"jsonrpc":"2.0","method":"anvil_dumpState","params":[]}' ${RPC_URL:-http://127.0.0.1:8545} | \
```
endersonmaia, line 8:
```suggestion
    --data '{"id":1,"jsonrpc":"2.0","method":"anvil_dumpState","params":[]}' "${RPC_URL:-http://127.0.0.1:8545}" | \
```

endersonmaia (changes requested):

.changeset/popular-needles-beam.md
endersonmaia, line 1: What does this filename mean?

endersonmaia (changes requested):

packages/anvil/package.json
tuler: I think so. This is the version for the anvil image. There is not much to add, or not to be called 1.0

PR #56, opened 2023-05-25
upload deb package to S3 with oclif

oap75:
We just need to set these GH Secrets : 

1.       GPG_PRIVATE_KEY
2.       AWS_ACCESS_KEY_ID
3.       AWS_SECRET_ACCESS_KEY
4.       GPG_PASSPHRASE
5.       IAM_ROLE_ARN
6.       AWS_REGION

The first secret is for `gpg key id` to create the `gpg` files, the second and third option is for `oclif IAM user` to upload to S3 in oclif upload section.

We need to create release.key with GPG and upload  to S3 bucket manually . 

image

### DIFF
--- a/.github/workflows/cli.yaml
+++ b/.github/workflows/cli.yaml
@@ -50,3 +50,8 @@ jobs:
               with:
                 name: Packages
                 path: ./apps/cli/dist/deb/*.deb
+
+            - name: Publish debian packages to S3
+              if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
+              run: oclif upload deb
+              working-directory: apps/cli
--- a/.github/workflows/cli.yaml
+++ b/.github/workflows/cli.yaml
@@ -34,3 +34,7 @@ jobs:
             - name: Build
               run: yarn build
               working-directory: apps/cli
+
+            - name: Install oclif
+              if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
+              run: yarn global add oclif
--- a/.github/workflows/cli.yaml
+++ b/.github/workflows/cli.yaml
@@ -39,17 +39,27 @@ jobs:
               if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
               run: yarn global add oclif
 
+            - name: Install GPG key for debian package signing
+              id: gpg_key
+              if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
+              uses: crazy-max/ghaction-import-gpg@v5
+              with:
+                  gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+                  passphrase: ${{ secrets.GPG_PASSPHRASE }}
+
             - name: Build debian package
               if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
               run: oclif pack deb
               working-directory: apps/cli
+              env:
+                  SUNODO_DEB_KEY: ${{ steps.gpg_key.outputs.keyid }}
 
             - name: Upload debian packages to Github artifacts
               if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
               uses: actions/upload-artifact@v3
               with:
-                name: Packages
-                path: ./apps/cli/dist/deb/*.deb
+                  name: Packages
+                  path: ./apps/cli/dist/deb/*.deb
 
             - name: Publish debian packages to S3
               if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
--- a/.github/workflows/cli.yaml
+++ b/.github/workflows/cli.yaml
@@ -43,3 +43,10 @@ jobs:
               if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
               run: oclif pack deb
               working-directory: apps/cli
+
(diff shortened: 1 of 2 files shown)


.github/workflows/cli.yaml
```diff
@@ -17,7 +17,12 @@ permissions:
     packages: write
 jobs:
     build:
+        name: release deb
```
tuler, line 20: Remove the name, because this build step will not be only for `deb`

.github/workflows/cli.yaml
```diff
@@ -17,7 +17,12 @@ permissions:
...
         runs-on: ubuntu-latest
+        env:
+          MYCLI_DEB_KEY: ${{ secrets.GH_MYCLI_DEB_KEY }}
```
tuler, line 23: rename to `SUNODO_DEB_KEY` and `secrets.SUNODO_DEB_KEY`

.github/workflows/cli.yaml
```diff
@@ -17,7 +17,12 @@ permissions:
...
+          AWS_ACCESS_KEY_ID: ${{ secrets.GH_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.GH_AWS_SECRET_ACCESS_KEY }}
```
endersonmaia, line 24: Where's the code where we create this credential/role?

It should be very restricted. 

The ideal is to use short-lived tokens, but if it's too complicated and the current role has the restrictions only to do what's needed by the job, I don't see a problem.

Maybe @tuler has another opinion on this.

tuler: The credential was being created by terraform. But I asked @oap75 to remove the creation of credentials, leaving the creation of a user, and I'd create credentials manually (using console), and copy to GitHub secret.

I'd rather use short lived. But it's complicated. Need that GitHub OIDC integration.

.github/workflows/cli.yaml
oap75: Fixed !

.github/workflows/cli.yaml
oap75: Fixed !

.github/workflows/cli.yaml
```diff
@@ -34,3 +39,24 @@ jobs:
             - name: Build
               run: yarn build
               working-directory: apps/cli
+
+            - name: Install oclif
+              if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
+              run: yarn global add oclif
+
+            - name: Build package
```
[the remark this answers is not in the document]
oap75: Fixed !

.github/workflows/cli.yaml
```diff
@@ -34,3 +39,24 @@ jobs:
...
+              working-directory: apps/cli
+
+            - name: Upload CLI to Github artifacts
+              if: startsWith(github.ref, 'refs/tags/@sunodo/cli@')
+              uses: actions/upload-artifact@v3
+              with:
+                name: Packages
+                path: ./apps/cli/dist/deb/*.deb
+
+            - name: Publish CLI to S3
```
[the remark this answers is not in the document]
oap75: Fixed !

.github/workflows/cli.yaml
oap75: I have created a separate [commit](https://github.com/sunodo/sunodo/pull/49/commits/00d261a830486ba3f4789228c30a0e4d7a359baa) to use terraform module to create the oclif user for us, if it's not needed, I can remove it .

apps/cli/package.json
```diff
@@ -66,6 +66,11 @@
             "@oclif/plugin-update"
         ],
         "topicSeparator": " ",
+        "update": {
+            "s3": {
+                "bucket": "cli-prod-sunodo-origin"
```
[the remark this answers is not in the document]
oap75: I couldn't find an option to use ENV. 

In the [Standalone tarballs](https://oclif.io/docs/releasing#standalone-tarballs) section, the doc says we need to set `oclif.update.s3.bucket` in `package.json` to a valid `S3 bucket` and have credentials set in `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment vars.

.github/workflows/cli.yaml
tuler: For reference, this is what I was talking about
https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services

endersonmaia: We use that strategy to run `cdk [diff|deploy]` via GHA on another private repository.

oap75: > https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services

we need to set credentials  in `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment vars The question is: 
  does the oclif work with this option ?

tuler: It does, because what it does it to generate short-lived credentials with AWS STS, and set the vars.

oap75: Fixed !

tuler (approved):

PR #71, opened 2023-05-29
Using retar tool of new SDK version

tuler:
This uses the retar tool of new SDK 0.1.0 (changed version strategy).


apps/cli/src/commands/build.ts
```diff
@@ -26,7 +26,7 @@ const CARTESI_LABEL_SDK_VERSION = `${CARTESI_LABEL_PREFIX}.sdk_version`;
 const CARTESI_LABEL_RAM_SIZE = `${CARTESI_LABEL_PREFIX}.ram_size`;
 const CARTESI_LABEL_DATA_SIZE = `${CARTESI_LABEL_PREFIX}.data_size`;
 
-const DEFAULT_SDK_VERSION = "0.15.0";
+const DEFAULT_SDK_VERSION = "0.1.0";
```
endersonmaia, line 29: Should we add a `SUNODO` or `SND` prefix to our globals/constants? just for disambiguation between `CARTESI`'s prefixes.

apps/cli/src/commands/build.ts
```diff
@@ -109,9 +109,9 @@ export default class BuildApplication extends Command {
             dataSize: labels[CARTESI_LABEL_DATA_SIZE] ?? "10Mb",
             entrypoint: imageInfo["Config"]["Entrypoint"] ?? [],
             env: imageInfo["Config"]["Env"] || [],
-            ramSize: labels[CARTESI_LABEL_RAM_SIZE] ?? DEFAULT_RAM_SIZE,
+            ramSize: labels[CARTESI_LABEL_RAM_SIZE] ?? CARTESI_DEFAULT_RAM_SIZE,
             sdkVersion:
-                labels[CARTESI_LABEL_SDK_VERSION] ?? DEFAULT_SDK_VERSION,
+                labels[CARTESI_LABEL_SDK_VERSION] ?? SUNODO_DEFAULT_SDK_VERSION,
```
endersonmaia, line 114: This is not gonna work great. `CARTESI_SDK_VERSION` and `SUNODO_SDK_VERSION` have different values.

If we wan't to label the image to have the information about which version of the CARTESI_SDK it's using, we should keep a it locked into `0.15.0`, and use the `SUNODO_SDK_VERSION`  to track which `sunodo/sdk` will be used.

endersonmaia (changes requested):

apps/cli/src/commands/build.ts
tuler: I think the label should be a sunodo label, not a cartesi label, because it tracks the sunodo sdk version.

endersonmaia (changes requested):
Squash the fixups.

We'll need a PR for `sunodo/sunodo-templates` compatible with the new `io.sunodo.sdk_version` label.

tuler:
https://github.com/sunodo/sunodo-templates/pull/9

endersonmaia (approved):

PR #74, opened 2023-05-29
Setting up gpg key for signing

tuler:
This uses the [import GPG key action](https://github.com/marketplace/actions/import-gpg) to import the key from GitHub secret to the keyring.

Then it uses the imported key ID in the debian package generation step.


oap75 (approved):
LGTM

PR #78, opened 2023-05-30
ci(cli): use existing OIDC provider

tuler:
Set `create_oidc_provider` to `false`, and use an existing OIDC provider I created manually.

### DIFF
--- a/apps/cli/tf/bucket.tf
+++ b/apps/cli/tf/bucket.tf
@@ -59,15 +59,13 @@ resource "aws_iam_policy" "sunodo_cli_s3_policy" {
 }
 
 module "github-oidc" {
-  source  = "terraform-module/github-oidc-provider/aws"
-
   # https://github.com/terraform-module/terraform-aws-github-oidc-provider
-  version = "~> 2.1.0"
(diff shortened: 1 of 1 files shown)


endersonmaia (approved):

apps/cli/tf/bucket.tf
```diff
@@ -59,15 +59,13 @@ resource "aws_iam_policy" "sunodo_cli_s3_policy" {
...
-  source  = "terraform-module/github-oidc-provider/aws"
-
   # https://github.com/terraform-module/terraform-aws-github-oidc-provider
-  version = "~> 2.1.0"
-  create_oidc_provider = true
-  create_oidc_role     = true
-
+  source                    = "terraform-module/github-oidc-provider/aws"
+  version                   = "~> 2.1.0"
+  oidc_provider_arn         = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:oidc-provider/token.actions.githubusercontent.com"
```
oap75, line 65: If we want to use terraform cloud, we could use variables in this section too . like `var.oidc_arn`.

PR #84, opened 2023-06-02
Contracts deployment packaging

tuler:
- downgrades target EVM of solidity code to `Paris` instead of `Shanghai` to keep anvil compatibility
- changes how deployment information is packaged.

Now there is another image called `sunodo/deployments` that holds the `localhost.json` deployment file and also the `anvil_state.json` file with the anvil state to be loaded.

The contracts information is printed when we run the image.

### DIFF
--- a/.github/workflows/contracts.yaml
+++ b/.github/workflows/contracts.yaml
@@ -1,0 +1,104 @@
+name: contracts
+on:
+    push:
+        tags:
+            - "@sunodo/contracts@*"
+    pull_request:
+        paths:
+            - .github/workflows/contracts.yaml
+            - packages/contracts/**
+concurrency:
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+permissions:
+    contents: read
+    packages: write
+jobs:
+    build:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v3
+
+            - name: Setup Node.js
+              uses: actions/setup-node@v3
+              with:
+                  node-version: 18
+                  cache: "yarn"
+                  registry-url: https://registry.npmjs.org
+
+            - name: Install Foundry
+              uses: foundry-rs/foundry-toolchain@v1
+
+            - name: Install Dependencies
+              run: yarn
+
+            - name: Build
+              run: yarn build
+              working-directory: packages/contracts
+
+            - name: Generate localhost deployment
+              run: ./dump_localhost.sh
+              working-directory: packages/contracts
+
+            - name: Docker meta
+              id: meta
+              uses: docker/metadata-action@v4
+              with:
+                  images: |
+                      docker.io/sunodo/deployments,enable=${{ github.event_name != 'pull_request' }}
+                      ghcr.io/sunodo/deployments
+                  tags: |
+                      type=match,pattern=(.+)@(.*),group=2
+                      type=ref,event=pr
+                  labels: |
+                      org.opencontainers.image.title=Sunodo's deployments
+                      org.opencontainers.image.description=Sunodo's deployments in an image
+
+            - name: Set up QEMU
+              uses: docker/setup-qemu-action@v2
+
+            - name: Set up Docker Buildx
+              uses: docker/setup-buildx-action@v2
+
+            - name: Login to GitHub Container Registry
+              uses: docker/login-action@v2
+              with:
+                  registry: ghcr.io
+                  username: ${{ github.actor }}
+                  password: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Login to DockerHub
+              uses: docker/login-action@v2
+              with:
+                  username: ${{ secrets.DOCKERHUB_USERNAME }}
+                  password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+            - name: Build and push
+              uses: docker/bake-action@v3
+              with:
+                  workdir: packages/contracts
+                  files: |
+                      ./docker-bake.hcl
+                      ./docker-bake.platforms.hcl
+                      ${{ steps.meta.outputs.bake-file }}
+                  set: |
+                      *.cache-from=type=gha
+                      *.cache-to=type=gha,mode=max
+                  push: true
+
+            - name: Publish package
+              if: startsWith(github.ref, 'refs/tags/@sunodo/contracts@')
+              run: npm publish --access public
+              env:
+                  NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+            - name: Upload to release
+              uses: softprops/action-gh-release@v1
+              if: startsWith(github.ref, 'refs/tags/@sunodo/contracts@')
+              with:
+                  body_path: packages/contracts/CHANGELOG.md
+                  files: |
+                      packages/contracts/anvil_state.json
+                      packages/contracts/deployments/**
+                      packages/contracts/export/abi/*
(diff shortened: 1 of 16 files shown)


tuler:
Tested and it works. The testing procedure is:

1. build `sunodo/deployments:devel` image:

```
yarn
cd packages/contracts
yarn build
./dump_localhost.sh
docker buildx bake --load
```

2. build `sunodo/rollups-node:devel`. Modify line 67 of `packages/rollups-node/Dockerfile`

```diff
-COPY --from=sunodo/deployments:0.1.0 /opt/cartesi/share/deployments/ /opt/cartesi/share/deployments/
+COPY --from=sunodo/deployments:devel /opt/cartesi/share/deployments/ /opt/cartesi/share/deployments/
```

```
cd packages/rollups-node
docker buildx bake --load
```

3. run a node for any dapp. Modify line 81 of `apps/cli/src/node/docker-compose.yml`:

```diff
-        image: sunodo/rollups-node:0.2.0
+        image: sunodo/rollups-node:devel
```

```
cd apps/cli
yarn build
cd <dapp_dir>
sunodo run -v
```

tuler:
The CI of rollups-node is not passing because I'm already locking to versioned images which don't exist yet.

packages/contracts/dump_localhost.sh
```diff
@@ -0,0 +1,23 @@
...
+rm -rf export/abi/localhost.json
+
+# deploy project to localhost (anvil)
+npx hardhat deploy --network localhost --export export/abi/localhost.json
+
+# kill anvil, gracefully, make it dump the state due to --dump-state
+kill $PID
+
+ls -al $STATE_FILE
+ls -al export/abi/
```
endersonmaia, line 23:
```suggestion
ls -al $STATE_FILE
ls -al export/abi/localhost.json
```

packages/contracts/dump_localhost.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# wait for anvil to start
+sleep 1
+
+# remove previous deployment
+rm -rf deployments/localhost
+rm -rf export/abi/localhost.json
+
+# deploy project to localhost (anvil)
+npx hardhat deploy --network localhost --export export/abi/localhost.json
```
endersonmaia, line 17: Does foundry have some tool to make this deployment?

The same as before, is it possible to use `cartesi/rollups-cli` for this?

.github/workflows/contracts.yaml
```diff
@@ -0,0 +1,104 @@
...
+              run: ./dump_localhost.sh
+              working-directory: packages/contracts
+
+            - name: Docker meta
+              id: meta
+              uses: docker/metadata-action@v4
+              with:
+                  images: |
+                      docker.io/sunodo/deployments,enable=${{ github.event_name != 'pull_request' }}
+                      ghcr.io/sunodo/deployments
```
endersonmaia, line 50: Should we name the package as deployments or the container image as contracts?

.github/workflows/contracts.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+            - name: Upload to release
+              uses: softprops/action-gh-release@v1
+              if: startsWith(github.ref, 'refs/tags/@sunodo/contracts@')
+              with:
+                  body_path: packages/contracts/CHANGELOG.md
+                  files: |
+                      packages/contracts/anvil_state.json
+                      packages/contracts/deployments/**
+                      packages/contracts/export/abi/*
```
endersonmaia, line 104: Where can I see what's the expected filename here, and how the files are organized inside?

packages/contracts/dump_localhost.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# kill anvil, gracefully, make it dump the state due to --dump-state
+kill $PID
```
endersonmaia, line 20:
```suggestion
kill "$PID"
```

packages/contracts/dump_localhost.sh
```diff
@@ -0,0 +1,23 @@
...
+
+ls -al $STATE_FILE
```
endersonmaia, line 22:
```suggestion
ls -al "$STATE_FILE"
```

packages/rollups-node/address_book
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
+if [ $# -eq 0 ]; then
+    echo "Usage: $0 <network | deployment_file>"
+    exit 1
+fi
+
+if test -f $1; then
```
endersonmaia, line 7:
```suggestion
if test -f "$1"; then
```

packages/rollups-node/address_book
```diff
@@ -0,0 +1,15 @@
...
+    FILE=$1
+else
+    NETWORK=$1
+    DIRECTORY=${DEPLOYMENTS_DIRETORY:-"/opt/cartesi/share/deployments"}
+    FILE="$DIRECTORY/$NETWORK.json"
+fi
+
+cat $FILE | jq -r '.contracts | to_entries[] | .key + ": " + .value.address'
```
endersonmaia, line 15:
```suggestion
cat "$FILE" | jq -r '.contracts | to_entries[] | .key + ": " + .value.address'
```

shellcheck suggested this, but I didn't test

```
cat "$FILE" | jq -r '.contracts | to_entries[] | .key + ": " + .value.address'
    ^-----^ SC2002 (style): Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.

For more information:
  https://www.shellcheck.net/wiki/SC2002 -- Useless cat. Consider 'cmd < file...

```

endersonmaia (changes requested):

packages/contracts/dump_localhost.sh
tuler: No, deploy through `forge` is still behind `hardhat`

.github/workflows/contracts.yaml
tuler: Problem is we've been using deployments all over, in cartesi

.github/workflows/contracts.yaml
tuler: there is one file per network, using the name of the network.
```
localhost.json
goerli.json
sepolia.json
...
```

.github/workflows/contracts.yaml
endersonmaia: So let's change the package name for deployments (`./packages/deployments`)

packages/contracts/dump_localhost.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env sh
+set -e
+STATE_FILE="anvil_state.json"
+
+# start anvil in the background, with --dump-state
+anvil --dump-state $STATE_FILE &
```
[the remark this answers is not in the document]
tuler: Actually this subproject has no README yet. Can we leave for another PR?
I'm pretty sure this package will need some more changes in a near future

.github/workflows/contracts.yaml
tuler: But the package is not deployments, it's contracts.
Deployments is a product of contracts.

endersonmaia:
squash the fixups

tuler:
If the tags `@sunodo/contracts@0.1.0` and `rollups-node@0.2.0` will be in the same commit, when we do the release, the CI still won't work.

The only way is to separate those two releases.

tuler:
I split this PR into 3.
This, https://github.com/sunodo/sunodo/pull/85 and https://github.com/sunodo/sunodo/pull/86

The idea is to merge and release this first, then the second, then finally the third.

endersonmaia (approved):
Since your're creating a new docker image, remember to add it to the images to me removed on the CI at https://github.com/sunodo/sunodo/blob/main/.github/workflows/rm-closed-pr-images.yaml#L15

PR #85, opened 2023-06-03
Using deployments docker image

tuler:
This PR makes use of the new image `sunodo/deployments:0.1.0`


endersonmaia (approved):

endersonmaia (approved):
I can't see the releases for the sunodo/deployments at GH Releases.

Approving since the images are already available on both container registries.

PR #92, opened 2023-06-06
Print endpoints

endersonmaia:


endersonmaia:
> Do you think it's necessary to have a separate "one shot" for that? Wouldn't be enough to be in the traefik one?

The idea was to just print the endpoints after traekif is ready, but I didn't implement traefik readiness verification.

I'll send a simpler approach.

endersonmaia:
@tuler you did `push --force` but there's a another commit (https://github.com/sunodo/sunodo/pull/92/commits/f7751a91542881ea0a15cea1c67b494bc920ca2d), and I don't know if it should be in this PR.

tuler:
Changed the base branch.
Please create a changeset

tuler:
change commit message to `feat(rollups-node)`

tuler (approved):

PR #95, opened 2023-06-12
Address book and send input commands

tuler:
This adds "two" commands:

- sunodo addres-book
- sunodo send

Address book prints the addresses of the smart contracts deployed to the local `anvil` service. This way the user can interact with those contracts, or write front-end application which interact with those contracts.

The send command and sub-commands provides a way to send inputs of several types to the application.

Running `sunodo send` (after running an application with `sunodo run`) presents the input types in a list.


apps/cli/src/node/docker-compose.yml
```diff
@@ -76,7 +76,7 @@ services:
             - ./.sunodo/image:/var/opt/cartesi/machine-snapshots/0_0:ro
 
     validator:
-        image: sunodo/rollups-node:0.2.0
+        image: sunodo/rollups-node:devel
```
endersonmaia, line 79: Should this be here?

apps/cli/src/wallet.ts
```diff
@@ -0,0 +1,138 @@
...
+import { mnemonicToAccount } from "viem/accounts";
+
+const DEFAULT_DEVNET_MNEMONIC =
+    "test test test test test test test test test test test junk";
+
+export interface EthereumPromptOptions {
+    supportedChains: Chain[];
+    chain?: string;
+    rpcUrl?: string;
+    // mnemonic?: string;
```
endersonmaia, line 21: Should this be commented? If not used, remove it.

tuler:
@SocketSecurity ignore @wagmi/cli@1.1.0
@SocketSecurity ignore viem@0.3.50

tuler:
The turbo cache hit of `yarn build --filter=contracts` is not working well.
For some reason the hit is not restoring the `build/anvil_state.json` from the cache, even thought the `build/**` is configured in `turbo.json` to be an output.

endersonmaia (approved):
LGTM

PR #109, opened 2023-06-19
ci: get docker tags from package.json

### LINKED ISSUE #105: Publishing strategy based on push of tag not working
## 🙂 Expected behavior

We designed a publishing strategy based on changesets where there is a [release workflow](https://github.com/sunodo/sunodo/blob/main/.github/workflows/release.yaml) that run on push to `main` branch and uses the [changesets action](https://github.com/changesets/action).

When there is a release, recognized by changesets, that executes [changesets tag](https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#tag) which creates git tags for all released packages, and those tags are pushed to the repo.

We expected that would trigger package specific workflows that would take care of publishing packages based on push of tag. Some packages are docker images, others are `npm` packages.

## 🫠 Actual behavior

The tags pushed by the release workflow don't trigger other workflows.
From [GitHub documentation](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow):

> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.

## 🧪 Minimal test case

It's hard to test these workflow in the real repo.
Maybe it's worth it to create a test repo and replicate part of our scenario there.

## ✔️ Possible solutions

1) ~~overcome the restriction by using a PAT instead of the GITHUB_TOKEN. I don't like using PAT because it becomes a management mess, and security burden.~~

2) use the fact that the above restriction does not apply to workflow dispatch.

- [ ] design the package specific workflow to also accept being triggered by workflow dispute
- [ ] change the release workflow to dispatch the other workflows
- [x] make the docker building workflows to use the version from the `package.json`, which is changed by changesets, instead of relying on push of tag (which wouldn't be triggered anymore)

endersonmaia:
With this PR, al docker packaged will still be built with the `:pr-NN` suffix.

Whenever a chageset release is made at the release workflow, and this released contains a docker package to build, the corresponding workflow will be called, instead of depending on the `on.push.tags` trigger.

/close #105

### DIFF
--- a/.github/workflows/anvil.yaml
+++ b/.github/workflows/anvil.yaml
@@ -1,8 +1,6 @@
 name: anvil
 on:
-    push:
-        tags:
-            - anvil@*
+    workflow_call:
     pull_request:
         paths:
             - .github/workflows/anvil.yaml
@@ -13,12 +11,19 @@ concurrency:
 permissions:
     contents: read
     packages: write
+    actions: write
 jobs:
     build:
         runs-on: ubuntu-latest
         steps:
             - name: Checkout
               uses: actions/checkout@v3
+
+            - name: Get package tag/version
+              id: package-version
+              if: ${{ github.event_name == 'push' }}
+              run: |
+                  jq -r '"PACKAGE_VERSION=\(.version)"' packages/anvil/package.json >> "$GITHUB_OUTPUT"
 
             - name: Docker meta
               id: meta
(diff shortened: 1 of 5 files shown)


tuler:
I don’t feel it’s safe to detect the version change based in the git history. I thought we would use the changesets action output to check if there is a package release, and which packages are them, and which version, on do a dispatch

endersonmaia:
The approach from the last fixup is using `on: workflow_run`. 

It works, but has the gotcha that every push to main will trigger the build for the other workflows even it there's not a release for them.

@tuler do you think it's acceptable?

I can work on another approach with less duplication, centralizing the docker build in a single docker-build parametrizable workflow, and trying to use the `published` and `publishedPackages` outputs from the `changeset/actions`. I'd need to experiment this approach yet.

tuler:
I think there was a misunderstanding.
I said the release.yaml workflow, which triggers on push to main, would dispatch the other docker workflows, conditionally based on the changeset action output, and setting the version as workflow input.

tuler (approved):

PR #111, opened 2023-07-03
Support NETWORK env var in build cmd

miltonjonat:
NETWORK is already an officially supported build arg in sunodo build, so it makes sense to allow it to also percolate to the back-end application

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -81,7 +81,7 @@ jobs:
     build_devnet:
         name: Build devnet
         needs: [release, packages_to_build]
-        if: ${{ needs.release.outputs.published == 'true' && contains(fromJSON(needs.packages_to_build.outputs.packages), 'devnet') }}
+        if: ${{ needs.release.outputs.published == 'true' && contains(fromJSON(needs.packages_to_build.outputs.packages), '@sunodo/devnet') }}
         uses: ./.github/workflows/devnet.yaml
         secrets: inherit
 


tuler (approved):

tuler:
Note that we are not storing local machine snapshots based on the build arguments, so this will not work well.

We should evaluate instead to remove the build arg, and see if we can live with a single machine build, especially now that smart contracts have the same address on all networks.

miltonjonat:
It's true that Cartesi Rollups' own contracts are now using deterministic deployments, and thus will have the same addresses for all supported networks. However, for DApps interacting with L1 contracts, it may still be the case that DApp logic is network-dependent.

On the other hand, maybe in these particular cases some other solution (such as a Relay contract) could be used instead.

tuler:
> for DApps interacting with L1 contracts, it may still be the case that DApp logic is network-dependent.

These L1 contracts can also use deterministic deployment, so the same applies.

miltonjonat:
Yeah, I'm certainly talking about 3rd party contracts, which are not under the control of the DApp developer.
For instance, has the "XYZ" ERC-20 of interest been deployed deterministically? I think CTSI itself hasn't. So if you want to accept only CTSI then it's already network dependent which address you need to check, right?

tuler:
> So if you want to accept only CTSI then it's already network dependent which address you need to check, right?

Right. In that case I'd do a configuration step post-deploy (a.k.a. setup input).

miltonjonat:
Right! And that's what I meant by using a Relay (a.k.a. setup input done right 😄) - in this case a `ChainIdRelay` or something like that.

felipeargento:
I believe it's unwise to commit to using the same address across different networks. While the property might seem great at first glance (and even that's up for debate), I'm skeptical about its long-term feasibility.

For Cartesi contracts, it's likely that different networks will have different contract addresses. This would be the case unless the chains are nearly identical. However, variables like block times, challenge periods, authority addresses, and others will vary from one chain to another, potentially affecting the behavior of the DApp. Although I don’t believe these factors would alter the DApp contract address directly, since it's somewhat isolated, it's still a point of consideration.

I wouldnt even advise the validator to use the same address for different networks. It might get confusing, you might fall victim to replay attacks etc. I'm in favor of things explicitly breaking if you're not a 100% sure you're interacting with the right chain!

(having said that, I like the env var...and think we should aim for the least amount of changes in code for deployment in different networks. Ideally it would all be automatic. Just talking about trying to maintain the addresses)

PR #118, opened 2023-07-26
fix: Changed /var/opt/cartesi/machine-snapshots permission to user ca…

lynoferraz:
…rtesi

### DIFF
--- a/apps/cli/src/node/docker-compose.yml
+++ b/apps/cli/src/node/docker-compose.yml
@@ -82,7 +82,7 @@ services:
                 condition: service_completed_successfully
         volumes:
             - blockchain-data:/opt/cartesi/share/deployments:ro
-            - ./.sunodo/image:/var/opt/cartesi/machine-snapshots/0_0
+            - ./.sunodo/image:/tmp/machine-snapshots/0_0:ro
         healthcheck:
             test: ["CMD", "is_ready"]
             interval: 10s
--- a/apps/cli/src/node/docker-compose.yml
+++ b/apps/cli/src/node/docker-compose.yml
@@ -76,7 +76,7 @@ services:
             - ./.sunodo/image:/var/opt/cartesi/machine-snapshots/0_0:ro
 
     validator:
-        image: sunodo/rollups-node:0.2.0
+        image: sunodo/rollups-node:0.3.0
         depends_on:
             dapp_deployer:
                 condition: service_completed_successfully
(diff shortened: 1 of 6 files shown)


packages/rollups-node/Dockerfile
```diff
@@ -68,7 +68,9 @@ COPY --from=inspect_server  /opt/cartesi/bin/inspect-server       /opt/cartesi/b
 COPY --from=graphql_server  /opt/cartesi/bin/graphql-server       /opt/cartesi/bin/graphql-server
 COPY --from=deployments     /opt/cartesi/share/deployments/       /opt/cartesi/share/deployments/
 
-RUN mkdir -p /var/opt/cartesi/machine-snapshots/0_0 && ln -s /var/opt/cartesi/machine-snapshots/0_0 /var/opt/cartesi/machine-snapshots/latest
+RUN mkdir -p /var/opt/cartesi/machine-snapshots/0_0 && \
+    ln -s /var/opt/cartesi/machine-snapshots/0_0 /var/opt/cartesi/machine-snapshots/latest && \
+    chown -R cartesi:cartesi /var/opt/cartesi/machine-snapshots
```
endersonmaia, line 73: Since we're breaking lines here, I suggest using the `<<EOF` style, to keep consistency in the file.

```suggestion
RUN <<EOF
mkdir -p /var/opt/cartesi/machine-snapshots/0_0
ln -s /var/opt/cartesi/machine-snapshots/0_0 /var/opt/cartesi/machine-snapshots/latest
chown -R cartesi:cartesi /var/opt/cartesi/machine-snapshots
EOF
```

endersonmaia (changes requested):
I suggest changing the commit message to something like that:

```
fix(rollups-node): change permissions

Change /var/opt/cartesi/machine-snapshots owner to cartesi:cartesi.
```

apps/cli/src/node/docker-compose.yml
```diff
@@ -82,7 +82,7 @@ services:
                 condition: service_completed_successfully
         volumes:
             - blockchain-data:/opt/cartesi/share/deployments:ro
-            - ./.sunodo/image:/var/opt/cartesi/machine-snapshots/0_0
+            - ./.sunodo/image:/tmp/machine-snapshots/0_0:ro
```
tuler, line 85: do we need `0_0` here?

endersonmaia: I wanted to be explicit. 

But it's not necessary and we change the up.sh script.

tuler (approved):

PR #119, opened 2023-07-27
AuthorityFactory and other deployment support contracts

tuler:
This PR deploys to Sepolia testnet the smart contracts that will support the DApp deployment system.
It includes:

- `AuthorityFactory`: a simple to use factory for Authorities. This will enable the indexing of Authorities and easy of configuration during dapp deployment by selecting the Authority from a list
- `ControlledDAppFactory`: a deployer of DApps that implements both `IFinancialProtocol` and `IMachineProtocol` and can be used for validators with no onchain payment.
- `ERC20DAppSystem`: a factory of deployer of DApps that can choose any ERC20 accepted as payment.


packages/contracts/contracts/deploy/ControlledDAppFactory.sol
endersonmaia, line 1: should the name be `Controlled,,,` or `Controller...`?

tuler: I meant `Controlled` due to the lack of a better name. It means the dapp execution is **controlled** by the validator.

packages/contracts/contracts/authority/AuthorityFactory.sol
```diff
@@ -0,0 +1,72 @@
...
+        authority.setHistory(history);
+        authority.transferOwnership(_authorityOwner);
+
+        emit AuthorityCreated(authority);
+
+        return authority;
+    }
+
+    function newAuthority() external override returns (Authority) {
+        return this.newAuthority(msg.sender);
```
guidanoli, line 27: Here, it is better to make the `newAuthority` you are calling a public function, and remove the `this.`. What this does is compile to a `JUMP` instead of a `CALL` opcode.

packages/contracts/contracts/authority/AuthorityFactory.sol
```diff
@@ -0,0 +1,72 @@
...
+        authority.setHistory(history);
+        authority.transferOwnership(_authorityOwner);
+
+        emit AuthorityCreated(authority);
+
+        return authority;
+    }
+
+    function newAuthority(bytes32 _salt) external override returns (Authority) {
+        return this.newAuthority(msg.sender, _salt);
```
guidanoli, line 46: Same here.

packages/contracts/contracts/authority/RoleBasedAuthority.sol
guidanoli, line 1: So, do you use this contract as the `IConsensus` a DApp subscribes to?
Because then, this contract must emit the `ApplicationJoined` event instead of the `Authority` within.
If you use the underlying `Authority` as the consensus, then it's ok.

tuler: I'm not using this contract. Maybe I should delete it?

packages/contracts/contracts/authority/AuthorityFactory.sol
tuler: But the function I'm calling is already a public function.
I just added this as a convenience option to remove one parameter.

packages/contracts/contracts/authority/RoleBasedAuthority.sol
guidanoli: If you're not using it, then I think it's ok to delete it.

packages/contracts/contracts/authority/AuthorityFactory.sol
guidanoli: By public, I mean it has the `public` modifier in its signature.
All `newAuthority` functions in this contract are external.
External functions can only be called externally (`CALL`).
Public functions can be called externally (`CALL`) and internally (`JUMP`).

packages/contracts/contracts/authority/AuthorityFactory.sol
guidanoli: By public, I mean it has the `public` modifier in its signature.
All `newAuthority` functions in this contract are external.
External functions can only be called externally (`CALL`).
Public functions can be called externally (`CALL`) and internally (`JUMP`).

packages/contracts/contracts/authority/AuthorityFactory.sol
tuler: So all `newAuthority` should be `public` instead of `external`?

guidanoli: Not all. Only those you're calling internally.

tuler: Done

packages/contracts/contracts/authority/RoleBasedAuthority.sol
tuler: Deleted

packages/contracts/contracts/authority/AuthorityFactory.sol
guidanoli: Now, you may remove the `this.` prefix.

```suggestion
        return newAuthority(msg.sender, _salt);
```

tuler:
Changes applied and rebased

guidanoli (approved):

PR #121, opened 2023-07-28
CLI for rollups 1.0.0

endersonmaia:


tuler (approved):

tuler:
rebased and added a few more changes

packages/machine-emulator-tools/README.md
```diff
@@ -7,15 +7,15 @@ It is currently distributed in binary format compiled for the `linux/riscv64` ar
...
+COPY --from=cartesi/machine-emulator-tools:0.12.0 / /
 ```
 
 While `cartesi` doesn't provide such image, Sunodo will provide the image under the `ghcr.io/sunodo` organization.
 
 So the application should include the following:
 
 ```dockerfile
-COPY --from=sunodo/machine-emulator-tools:0.11.0-ubuntu22.04 / /
+COPY --from=sunodo/machine-emulator-tools:0.12.0-ubuntu22.04 / /
```
endersonmaia, line 18: is it still `ubuntu22:04` ?

packages/machine-emulator-tools/Dockerfile
```diff
@@ -6,6 +6,5 @@ ADD https://github.com/cartesi/machine-emulator-tools/releases/download/v${VERSI
 RUN mkdir -p /tmp/machine-emulator-tools && tar -C /tmp/machine-emulator-tools -xzf /tmp/machine-emulator-tools.tar.gz
 
 FROM scratch as tools
-COPY --from=download /tmp/machine-emulator-tools/sbin/init /usr/sbin/init
 COPY --from=download /tmp/machine-emulator-tools/etc/hostname /etc/hostname
 COPY --from=download /tmp/machine-emulator-tools/opt/cartesi /opt/cartesi/
```
endersonmaia, line 10: looks like `/opt/cartesi` is being deprecated in favor os system's standard paths

should we change this to keep it consistent?

packages/rollups-node/Dockerfile
```diff
@@ -60,13 +58,12 @@ cp "$tmp_dir"/traefik /usr/local/bin/traefik
...
-COPY --from=graphql_server  /opt/cartesi/bin/graphql-server       /opt/cartesi/bin/graphql-server
-COPY --from=deployments     /opt/cartesi/share/deployments/       /opt/cartesi/share/deployments/
+COPY --from=state_server    /opt/cartesi/bin/cartesi-rollups-state-server   /usr/bin/cartesi-rollups-state-server
+COPY --from=indexer         /opt/cartesi/bin/cartesi-rollups-indexer        /usr/bin/cartesi-rollups-indexer
+COPY --from=dispatcher      /opt/cartesi/bin/cartesi-rollups-dispatcher     /usr/bin/cartesi-rollups-dispatcher
+COPY --from=advance_runner  /opt/cartesi/bin/cartesi-rollups-advance-runner /usr/bin/cartesi-rollups-advance-runner
+COPY --from=inspect_server  /opt/cartesi/bin/cartesi-rollups-inspect-server /usr/bin/cartesi-rollups-inspect-server
+COPY --from=graphql_server  /opt/cartesi/bin/cartesi-rollups-graphql-server /usr/bin/cartesi-rollups-graphql-server
 
 RUN mkdir -p /var/opt/cartesi/machine-snapshots/0_0 && ln -s /var/opt/cartesi/machine-snapshots/0_0 /var/opt/cartesi/machine-snapshots/latest
```
endersonmaia, line 68: same here, should we change from `/var/opt/cartesi/` to `/var/cartesi` ?

endersonmaia:
@tuler the `@sunodo/devnet`  package should release a docker AND an npm package, is that it?

If that's the case, I miss the CI that builds and releases the docker image.

endersonmaia:
@tuler still on the `devnet` package, locally I can't build it against the `sunodo/anvil:devel`.

when running `yarn build` locally, we should target all `:devel` docker images, and when building at CI we should target the latest released versions. 

Would that be possible? Is that desirable?

OR, the other way around, the default is to target final releases, and locally I should override it with a `:devel` argument or something.

packages/machine-emulator-tools/Dockerfile
tuler: This is still like it in the tar.gz.
Actually we need to change this to just copy everything to `/`

tuler:
> @tuler the `@sunodo/devnet` package should release a docker AND an npm package, is that it?
> 
> If that's the case, I miss the CI that builds and releases the docker image.

Just a docker. But I called it @sunodo/devnet anyway, I think it’s more consistent, even if it’s not a npm package.

I’ll work on the CI for it.

packages/rollups-node/Dockerfile
endersonmaia: Yes, we were doing the "right" thing, but people have different opinions on what's the right thing.

Actually, there is more than one way of doing it right. :)

But let us be consistent between projects.

tuler (approved):

PR #125, opened 2023-08-02
Display account balance in deploy interactive mode

tuler:
During account selection in deploy interactive mode we query the balance of each account to make it easier to choose.


endersonmaia:
@tuler I noticed that in this branch, the deployment with other accounts. other than the first one, doesn't fail.

tuler:
Yes, I fixed the bug ;-)

PR #127, opened 2023-08-02
WalletConnect not working

tuler:
Wallet Connect complains about a missing required dependency when we try to use it.


endersonmaia (approved):
I can´t properly test WalletConnect.

PR #129, opened 2023-08-04
Payable Factory

### LINKED ISSUE #128: Controller with manual control
## 📚 Context

The cartesi rollups node is a component that runs a node for a single specific application.
The sunodo controller is a component that observes the blockchain for deployed applications and spawns cartesi rollups nodes for each application.

The controller, after observing a deploy on the blockchain, must decide whether it will launch a node for it or not.

The first criteria is based on whether the wallet of the controller (which will send claims of the validation) matches the owner of the authority of the dapp.

The second criteria is cost related, whether the controller has the financial incentives to run a node for the application or not. There are two methods of implementing the financial incentive:

1. an onchain incentive using ERC-20 or any other financial mechanism of the blockchain.
2. on offchain incentive, where the controller will watch some outside integration, or be manually controlled.

The [ControlledDAppFactory](https://github.com/sunodo/sunodo/blob/main/packages/contracts/contracts/deploy/ControlledDAppFactory.sol) smart contract has the intention of implementing mechanism number 2 above.

However it fails to do that because it controls the `start` and `stop` methods using a `CONTROLLER_ROLE` granted in the constructor. As of now it grants control to `msg.sender` of the deployer of the contract, which is wrong because we are using deterministic deployment and the `msg.sender` in that case is the deploy proxy smart contract, so we have really no control.

Even if we fix the `CONTROLLER_ROLE` to grant it to the "correct" actor, the question of who is the correct actor is still pending. The intention was that the "authority" would be control, but that is not what is implemented.

## ✔️ Solution

One solution is to have one dapp factory for each authority. This factory would wire the authority automatically for one specific authority.

Another option would go another route and don't use the blockchain for this control, and expose an endpoint in the controller that starts and stops nodes. Then we could implement another component that interacts with that endpoint. Security is a concern in that case, the API must be authenticated somehow.

tuler:
This simplifies the deployment by encapsulating the consensus into the dapp factory.
Instead of the user choosing a consensus he will choose the factory to use.

Each authority will have his own factory, which will allow him to have more control over the deployment and the billing.
I deleted the ControlledDAppFactory, and every deploy will go through a PayableDAppFactory which accepts ERC-20.
The price is dictated by factory owner, and cannot be modified.

The factory also allows to import a dapp, previously instantiated, to be managed by the factory (in terms of billing). I require that the dapp has the same consensus as the consensus of the factory.

closes #128


packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /// @dev This function deploys a new application, and registers it with the payment system
+    function newApplication(
+        address _dappOwner,
+        bytes32 _templateHash,
+        string memory _cid,
+        uint256 _runway
+    ) external override returns (CartesiDApp) {
+        // delegate call to factory, owner must be this contract, so we can register with payment system
```
felipeargento, line 57: Isn't it cheaper to create the DApp with the correct owner and use another method to register it with the payment system? Something like an event.

Might be worth testing :)

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,132 @@
...
+        dapp.transferOwnership(_dappOwner);
+
+        // set initial runway
+        this.extendRunway(dapp, _runway);
+
+        // return application
+        return dapp;
+    }
+
+    function importApplication(
```
felipeargento, line 75: I think this should only be callable by the DApp owner! Or it should check if this application was already important. I'm worried about anyone being able to call this repeatedly and change the machine location.

(though `extendRunway` should be callable by anyone)

felipeargento (commented):
There is some room for gas optimization, but I think that is secondary/should be done after tests are written!

packages/contracts/contracts/deploy/PayableDAppFactory.sol
tuler: Are you talking about the following `transferOwnership`?
I think we can get rid of the `transferOwnership` and use `_dappOwner` right away, right?

felipeargento: Yeah! Sorry, marked the wrong line

packages/contracts/contracts/deploy/PayableDAppFactory.sol
tuler: True. Beware that the CID if content addressable, so if it's a different CID it's a different machine, and consequently a different machine hash. So when the validator try to start the node it will fail because it won't match the onchain hash.

tuler:
@felipeargento what do you think about not having the manual control anymore, for offchain payments?
I could add an exception in the `cost` function to say it's zero for the factory owner. So he could receive the money offchain, and extend at zero cost.

(does it make sense to have logic on a view function that depends on `msg.sender`?)

If the factory owner is also a minter of the token used, that is fine, he receives money offchain, and mint tokens to extend the runway, without changing any logic.

But if the token is say USDC, he will have to on-ramp fiat to USDC, to extend, which is also fine I guess.

tuler:
@felipeargento I fixed the problem with the `msg.sender` with [this change](https://github.com/sunodo/sunodo/pull/129/commits/3a3b661799305c3d5d64857b63c0ad23c5a5f044).

tuler:
Do you think it makes sense to add [Pausable](https://docs.openzeppelin.com/contracts/4.x/api/security#Pausable) to the factory?

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,144 @@
...
+        uint256 _time
+    ) external returns (uint256) {
+        return _extendRunway(msg.sender, _dapp, _time);
+    }
+
+    function _extendRunway(
+        address _msgSender,
+        ICartesiDApp _dapp,
+        uint256 _time
+    ) private returns (uint256) {
```
felipeargento, line 111: Isn't it better for this to be internal and not private? In case you want to inherit this contract in another one and use it.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,144 @@
...
+        // calculate the token cost of the runway time
+        uint256 _cost = this.cost(_time);
```
felipeargento, line 113: I think it would be better to classify the function `cost` as `public` and just call it here, rather than using `this` (which generates a call to address(this))

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,144 @@
...
+
+        // if runway was in the past, update to now
+        currentRunway = currentRunway < block.timestamp
+            ? block.timestamp
+            : currentRunway;
+
+        // update dapp runway
+        uint256 newRunway = currentRunway + _time;
+        runway[_dapp] = newRunway;
+
```
felipeargento, line 131: Here would be a good place to have the maximum runway check.
Something like:  `require( newRunway <= block.timestamp + 1 month, "can't extend until more than one month in the future")`.

I don't know what the conclusion was, but I think in our conversation we were thinking that it made sense to limit the max runway

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,144 @@
...
+        // emit event so offchain gets notified
+        emit FinancialRunway(address(_dapp), newRunway);
+
+        return newRunway;
+    }
+
+    function withdraw(uint256 amount) external onlyOwner {
```
felipeargento, line 138: Have you thought about making it `withdraw()` without an amount and always withdraw the max (`balanceOf(this)`)? I think its a more elegant way to do it, it's usually what you're going for when withdrawing. And makes the order of this transaction less important in the scenario in which someone deposited money right before you're `balanceOf` check and your `withdraw(balancOf)` call...which is not much, but its something

felipeargento:
> Do you think it makes sense to add [Pausable](https://docs.openzeppelin.com/contracts/4.x/api/security#Pausable) to the factory to pause all operations (`newApplication`, `importApplication` and `extendRunway`)

I think it makes total sense to make it Pausable!

Just making sure, cause I think we talked about this in person:
If an owner wants to increase prices, is the plan to deploy a new factory and point everyone there. Right?
Because I think there might be an easier solution...the way you're updating the runway, just changing the price shouldn't impact those who already extended runway with the old price. So just making the price not `immutable` and adding a function `onlyOwner` to change it should work fine.

I looked at the smart contracts only, but they're looking good! After they're fully tested we can look for gas optimization and possible attacks :)

packages/contracts/contracts/deploy/PayableDAppFactory.sol
tuler: We talked about that because the customer can't withdraw.
The validator can withdraw and refund the customer though, in a friendly relationship.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
tuler: Do you think we need both methods? Or just the `withdraw()`?
The simpler the better IMO.

tuler:
> I think it makes total sense to make it Pausable!
Ok

> the way you're updating the runway, just changing the price shouldn't impact those who already extended runway with the old price.

Yeah, I considered that. But it might be weird as a sales strategy to advertise a very low price, and then raising it all of a sudden, and forcing customers to migrate to another service. A fixed price is easier to understand, and to implement, and less prone to errors. In the future we could create more complex plans.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
felipeargento: Agreed! I think just `withdraw()` is better. 
Also saves you from having to count the 18 zeros for decimals haha

packages/contracts/contracts/deploy/PayableDAppFactory.sol
felipeargento: True.  However, I think it might make accounting harder offchain. You never know whats the amount of funds that you might have to give back. I also think it makes the price raising thing a bit more complicated, even if you price raise by migrating.

But I see your point!

tuler:
Added Pausable

felipeargento (approved):
Smart contracts look good to me!

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,155 @@
...
+    Pausable
+{
+    ICartesiDAppFactory public immutable factory;
+    IERC20 public immutable token;
+    IConsensus public immutable consensus;
+    uint256 public immutable price;
+
+    mapping(ICartesiDApp => uint256) public runway;
+
+    /// @notice Construct a new ERC20DAppFactory with the given token as payment method
```
guidanoli, line 33: Shouldn't it be `PayableAppFactory` here?

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,155 @@
...
+    /// @param _factory The factory to delegate calls to instantiate the dapp
```
guidanoli, line 34: The documentation for parameters `_owner` and `_token` are missing.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,155 @@
...
+        emit MachineLocation(address(dapp), _cid);
+
+        // set initial runway
+        _extendRunway(msg.sender, dapp, _runway);
+
+        // return application
+        return dapp;
+    }
+
+    function importApplication(
```
guidanoli, line 77: What if `_dapp` was not created by `factory`? Would this be a problem?

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,155 @@
...
+        token = _token;
+        consensus = _consensus;
+        price = _price;
+
+        // transfer ownership to the give owner
+        transferOwnership(_owner);
+    }
+
+    /// @dev This function deploys a new application, and registers it with the payment system
+    function newApplication(
```
guidanoli, line 54: Do you plan to add an entry point for deterministic deployment as well?

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,155 @@
...
+        ICartesiDApp _dapp,
+        uint256 _time
+    ) internal returns (uint256) {
+        // calculate the token cost of the runway time
+        uint256 _cost = cost(_time);
+
+        // transfer tokens from sender to this contract
+        require(
+            token.transferFrom(_msgSender, address(this), _cost),
+            "PaymentDAppFactory: failed to transfer tokens"
```
guidanoli, line 121: You may save some gas on deployment if you manage to shorten this string to less than 32 characters.
Alternatively, you can raise a [custom error](https://soliditylang.org/blog/2021/04/21/custom-errors/) and save even more gas.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,155 @@
...
+        // emit event so offchain gets notified
+        emit FinancialRunway(address(_dapp), newRunway);
+
+        return newRunway;
+    }
+
+    function withdraw() external onlyOwner {
+        require(
+            token.transfer(msg.sender, token.balanceOf(address(this))),
+            "PaymentDAppFactory: failed to withdraw tokens"
```
guidanoli, line 144: The same comment about error message size (or custom error) applies here.

guidanoli (commented):
I know this has been merged already, but I noticed how we could improve some things in the contracts.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
tuler: Do you see any problem?

I'm allowing only the dapp owner to do it.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
tuler: I don't see a need now

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,155 @@
...
+        return dapp;
+    }
+
+    function importApplication(
+        CartesiDApp _dapp,
+        string memory _cid,
+        uint256 _runway
+    ) external override whenNotPaused {
+        require(
+            _dapp.getConsensus() == this.consensus(),
```
guidanoli, line 83: Why not read `consensus` directly from storage instead of making an external call?

packages/contracts/contracts/deploy/PayableDAppFactory.sol
guidanoli: What if the DApp contract is malicious?

packages/contracts/contracts/deploy/PayableDAppFactory.sol
tuler: How?

packages/contracts/contracts/deploy/IPayableDAppSystem.sol
```diff
@@ -0,0 +1,31 @@
...
+        IConsensus consensus,
+        uint256 price
+    );
+
+    /// @notice Create a new ERC20 based DApp factory using the specified token
+    function newPayableDAppFactory(
+        IERC20 _token,
+        IConsensus consensus,
+        uint256 _price
+    ) external returns (IPayableDAppFactory);
```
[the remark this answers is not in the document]
guidanoli: Then, you'd define a new interface for it.
I like to type factories with the actual contracts instead of their interfaces, because then you generally have more control over the contracts that you've just deployed.
For example, imagine a factory that deploys `Ownable` contracts. If you type them by the interface, they will not have the functions from `Owanble` such as `transferOwnership`, which might be useful for factories that use other factories.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
guidanoli: `consensus` instead of `this.consensus()`.

packages/contracts/contracts/deploy/IPayableDAppSystem.sol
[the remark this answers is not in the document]
tuler: `Ownable` is an abstract contract, not an interface.
If there was an interface, i.e. `IOwnable` I'd use it in my `IOwnableFactory`, and `transferOwnership` would be part of the interface rather than the contract.

Do you have another example?

I still think the correct is to return the interface.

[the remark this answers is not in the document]
tuler: > you generally have more control over the contracts that you've just deployed

It shouldn't matter, as long as it implements the interface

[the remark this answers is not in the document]
guidanoli: > If there was an interface, i.e. IOwnable I'd use it in my IOwnableFactory, and transferOwnership would be part of the interface rather than the contract.

The thing is, that there isn't such an interface.
OpenZeppelin only defines `Ownable`, but no `IOwnable`.
That is why I chose to make `CartesiDAppFactory` return `CartesiDApp`.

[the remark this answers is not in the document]
guidanoli: > It shouldn't matter, as long as it implements the interface

It matters, if the contract contains functions that are not exposed by the interface.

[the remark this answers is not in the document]
tuler: > It matters, if the contract contains functions that are not exposed by the interface.

The interface should expose all functions that make sense for the interface.

If you need a function of the contract you type cast it manually and use it.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
tuler: Can you elaborate?

packages/contracts/contracts/deploy/IPayableDAppSystem.sol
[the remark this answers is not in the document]
guidanoli: > The case of OpenZeppelin is not related to the case of CartesiDAppFactory

It is, because `CartesiDApp` inherits from `Ownable`.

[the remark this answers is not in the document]
guidanoli: > The interface should expose all functions that make sense for the interface.

That is not always the case. Some contracts inherit from several interfaces.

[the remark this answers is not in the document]
tuler: > Some contracts inherit from several interfaces.

True, so you cast to the interface you want to interact as.

> It is, because CartesiDApp inherits from Ownable.

So if you want a `ICartesiDApp` to always behave as an `Ownable` you should create a `transferOwnership` method in the `ICartesiDApp`.


What defines expected behavior is the interface, not the implementation of the interface.
That is the whole point of interfaces. Clients should be dealing with interfaces, not concrete classes.

That is how interfaces in general programming languages work, and I don't see the reason to be different in solidity. Maybe I'm missing something.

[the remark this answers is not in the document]
guidanoli: > True, so you cast to the interface you want to interact as.

Yes, Solidity allows you to cast any address value to any contract type. But I find it weird that you're upcasting `Apple` to `Fruit`, exposing `Fruit` and then downcasting to `Apple` again. Why not expose `Apple` instead?

> So if you want a ICartesiDApp to always behave as an Ownable you should create a transferOwnership method in the ICartesiDApp.

The thing is that you can't, with the current state of Solidity. The compiler raises an error saying that the function has already been defined (in this case, in `Ownable`).

> Clients should be dealing with interfaces, not concrete classes.

I generally agree with this statement. But factories are a special case, because their whole job is to deploy actual, concrete contracts.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -0,0 +1,155 @@
...
+            "PaymentDAppFactory: cannot import dapp with a different consensus"
+        );
```
guidanoli, line 85: It is worth noting that this assertion might only work as a safety check for the inattentive DApp owner to migrate to the correct consensus. The DApp owner can, afterward, migrate to any consensus they so desire.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
guidanoli: Never mind, malicious DApp implementation couldn't any harm.

packages/contracts/contracts/deploy/IPayableDAppSystem.sol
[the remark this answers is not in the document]
tuler: > I find it weird that you're upcasting `Apple` to `Fruit`

I'm not upcasting, I'm just returning the general representation

> and then downcasting to `Apple` again.

I won't downcast to `Apple`.

> The thing is that you can't, with the current state of Solidity. The compiler raises an error saying that the function has already been defined (in this case, in `Ownable`).

This is unfortunate.

packages/contracts/contracts/deploy/PayableDAppFactory.sol
tuler: This contract is just managing funding.

PR #138, opened 2023-08-16
Query list of factories instead of list of authorities

tuler:
We already have indexing of factories on sepolia, so we can adjust the implementation that queries the list while deploying an application

### DIFF
--- a/apps/cli/graphql/queries.graphql
+++ b/apps/cli/graphql/queries.graphql
@@ -3,3 +3,16 @@ query authorities {
         id
     }
 }
+
+query factories {
+    factories {
+        id
+        price
+        token {
+            decimals
+            id
+            name
+            symbol
+        }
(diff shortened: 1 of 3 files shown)


endersonmaia:
@tuler I need some explanations to understand what this PRs does.

tuler:
This PR adds the query of factories from the subsquid we implemented in the PR #135
The subsquid is deployed to https://squid.subsquid.io/sunodo-sepolia/graphql
The query we need to do is:

```graphql
query MyQuery {
  factories {
    id
    price
    token {
      decimals
      id
      name
      symbol
    }
  }
}
```

This brings the factories, and for each factory the price and token used by the factory.
Then we can build a list for the user selection.

We can still improve the UI by also showing the price next to the factory address, like `12.5 SUN/week`

PR #150, opened 2023-08-18
Revenue share model

tuler:
The idea of this PR is to implement a revenue share model between the `PayableDAppFactory` and the `PayableDAppSystem` (which creates the first).

The implementation uses the OpenZeppelin [PaymentSplitter](https://docs.openzeppelin.com/contracts/4.x/api/finance#PaymentSplitter) to share payments between the two.

The share % would be defined by the owner of the `PayableDAppSystem`.

I'm still figuring out how to define the address of the payee of the `PayableDAppFactory` counterpart. The `PaymentSplitter` contract is immutable, so it's not possible to change the payees, so any change would need a new instance.

### DIFF
--- a/.changeset/grumpy-rice-hide.md
+++ b/.changeset/grumpy-rice-hide.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/contracts": minor
+---
+
+add a revenue share model to Marketplace
--- a/packages/contracts/contracts/marketplace/Marketplace.sol
+++ b/packages/contracts/contracts/marketplace/Marketplace.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.8.13;
 
 import {ICartesiDAppFactory} from "@cartesi/rollups/contracts/dapp/ICartesiDAppFactory.sol";
 import {IConsensus} from "@cartesi/rollups/contracts/consensus/IConsensus.sol";
+import {PaymentSplitter} from "@openzeppelin/contracts/finance/PaymentSplitter.sol";
 import {ENS} from "@ensdomains/ens-contracts/contracts/registry/ENS.sol";
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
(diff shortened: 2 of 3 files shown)


packages/contracts/contracts/deploy/PayableDAppSystem.sol
```diff
@@ -43,4 +59,16 @@ contract PayableDAppSystem is IPayableDAppSystem {
 
         return payableFactory;
     }
+
+    function _createPaymentSplitter(
+        address _payee
```
guidanoli, line 64: This parameter `_payee` can be confused with `payee`.
Should we rename it? `_otherPayee`?

guidanoli:
Could you give me an example where this revenue share model would come in handy?

tuler:
> 

I wrote an issue with more details at #153

packages/contracts/contracts/deploy/PayableDAppSystem.sol
```diff
@@ -16,9 +17,25 @@ contract PayableDAppSystem is IPayableDAppSystem {
...
+    address public immutable payee;
+
+    constructor(
+        ENS _ens,
+        ICartesiDAppFactory _factory,
+        uint256 _systemShares,
+        address _payee
+    ) {
+        require(
+            _systemShares < TOTAL_SHARES,
```
guidanoli, line 31: Why not `<=`? The system wouldn't break if the factory payee had no shares, would it?

packages/contracts/contracts/deploy/PayableDAppSystem.sol
```diff
@@ -16,9 +17,25 @@ contract PayableDAppSystem is IPayableDAppSystem {
...
+            "PayableDAppSystem: systemShares must be lower than TOTAL_SHARES"
```
guidanoli, line 32: We typically say `<` is "less than", not "lower than".

packages/contracts/contracts/deploy/PayableDAppSystem.sol
```diff
@@ -16,9 +17,25 @@ contract PayableDAppSystem is IPayableDAppSystem {
     ENS public immutable ens;
     ICartesiDAppFactory public immutable factory;
 
-    constructor(ENS _ens, ICartesiDAppFactory _factory) {
+    uint256 public constant TOTAL_SHARES = 10000;
```
guidanoli, line 20: Why not make `TOTAL_SHARES` a constructor argument, like `systemShares`?
Say you want to have one third of the shares for the factory.
With `TOTAL_SHARES` being set to `10000`, you're prone to rounding errors ($.3333 \ne \frac{1}{3}$).

PR #151, opened 2023-08-21
feat: add workflow status badge

### LINKED ISSUE #134: Workflow status badge
## 📚 Context

The repo doesn't have A status badge. It shows whether a workflow is currently failing or passing. It can be useful to show the latest status of the GH actions. 

## ✔️ Solution

A common place to add a status badge is in the README.md file of the repository.

https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge

## 📈 Subtasks

-   [ ] define badge for important actions

## 🎯 Definition of Done

Sample repo : https://github.com/elbruno/AzureApiBadge/tree/master

oap75:
/close #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Sunodo
 
+![Release Packages](https://github.com/sunodo/sunodo/actions/workflows/release.yaml/badge.svg)
+
 Sunodo is a framework for developing decentralized applications built on top of [Cartesi](http://cartesi.io) technology.
 It helps developers to:
 


endersonmaia:
Since this is a monorepo and we have a lot of projects, I don't think this is enough.

If we want a badge, we should have one for each package's README.md AND/OR all packages' badges in the main /README.md.

What do you think?

oap75:
> Since this is a monorepo and we have a lot of projects, I don't think this is enough.
> 
> If we want a badge, we should have one for each package's README.md AND/OR all packages' badges in the main /README.md.
> 
> What do you think?

In [release.yaml](https://github.com/sunodo/sunodo/blob/f7d286b2c6f92a2cf28c4325af736265e2a2aaab/.github/workflows/release.yaml), we call all other actions to build the packages. so, I think we should show the status of this workflow on the main README and I agree with U. we can show the status of each workflow to it's related README too.

endersonmaia (approved):

tuler (approved):

PR #161, opened 2023-08-28
ENS setName support

tuler:
This allows the `PayableDAppFactory` to be named by its owner.

ENS requires that the reverse registration of a contract (given an address resolve a name) that the transaction must come from the address being registered.

This is implemented by a method `setName` callable by the `PayableDAppFactory` owner.

### DIFF
--- a/packages/contracts/contracts/deploy/IPayableDAppFactory.sol
+++ b/packages/contracts/contracts/deploy/IPayableDAppFactory.sol
@@ -18,7 +18,7 @@ interface IPayableDAppFactory {
     /// @param _dappOwner The address that should own the DApp
     /// @param _templateHash The hash of the initial state of the Cartesi Machine
     /// @param _cid The IPFS CID of the cartesi machine snapshot
-    /// @param _runway the amount of time in seconds the dapp will be running (will trigger transfer)
+    /// @param _runway The amount of time in seconds the dapp will be running (will trigger transfer)
     /// @return The application
     function newApplication(
         address _dappOwner,
@@ -30,12 +30,16 @@ interface IPayableDAppFactory {
     /// @notice Import an existing dapp to be managed by this factory
     /// @param _dapp The dapp to be imported
     /// @param _cid The IPFS CID of the cartesi machine snapshot
-    /// @param _runway the amount of time in seconds the dapp will be running (will trigger transfer)
+    /// @param _runway The amount of time in seconds the dapp will be running (will trigger transfer)
     function importApplication(
         CartesiDApp _dapp,
         string memory _cid,
         uint256 _runway
     ) external;
+
+    /// @notice Sets a name for the factory using ENS service
+    /// @param _name The name of the factory
+    function setName(string memory _name) external;
 
     /// @notice Returns the cost of running a DApp for a given time
     /// @param _time The time in seconds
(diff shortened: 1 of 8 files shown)


packages/contracts/contracts/deploy/IPayableDAppFactory.sol
```diff
@@ -37,6 +37,10 @@ interface IPayableDAppFactory {
         uint256 _runway
     ) external;
 
+    /// @notice sets a name for the factory using ENS service
+    /// @param _name name of the factory
```
guidanoli, line 41: We usually capitalize the first word after `@<tag>`, because, if this text gets scrapped to generate documentation, it will be converted into an HTML paragraph. Also, we generally use definite articles to describe parameters.

```suggestion
    /// @notice Sets a name for the factory using ENS service
    /// @param _name The name of the factory
```

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -50,6 +56,16 @@ contract PayableDAppFactory is
         transferOwnership(_owner);
     }
 
+    /// @notice sets a name for the pool using ENS service
```
guidanoli, line 59: This comment can be removed, since it will inherit from the interface, if you ever use some tool to scrape the documentation off these contracts. Also, why "pool"?

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -50,6 +56,16 @@ contract PayableDAppFactory is
...
+    function setName(string memory name) external override onlyOwner {
+        ReverseRegistrar ensReverseRegistrar = ReverseRegistrar(
```
guidanoli, line 61: I think it's better to cast it as `IReverseRegistrar`, since we're not really concerned about the implementation, right?

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -50,6 +56,16 @@ contract PayableDAppFactory is
...
+            ens.owner(ADDR_REVERSE_NODE)
+        );
+
+        // call the ENS reverse registrar resolving address to name
```
guidanoli, line 65: I think "to resolve" sounds better, no?

```suggestion
        // call the ENS reverse registrar to resolve address to name
```

tuler: Actually I think "to register name of address" makes more sense.

guidanoli: Yes, this works as well.

guidanoli (approved):

packages/contracts/deploy/02_factory.ts
```diff
@@ -1,23 +1,35 @@
...
+    const ens = await deployENS(deployer);
+    await ens.createTopLevelDomain("test");
+    return ens.ens.address;
+};
 
 const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
-    const { deployments, getNamedAccounts } = hre;
+    const { deployments, ethers, getNamedAccounts } = hre;
     const { deployer } = await getNamedAccounts();
+    const [deployer_] = await ethers.getSigners();
```
guidanoli, line 15: These names can be confusing. What is the difference between `deployer` and `deployer_`?

tuler: Renamed the variable to `deployerSigner`

guidanoli (approved):

PR #162, opened 2023-08-28
Specify payee to PayableDAppFactory

tuler:
This specifies the payee of a `PayableDAppFactory`, instead of depositing the money in the contract itself.
The transfer is now from the deployer to the payee.
The withdraw method was removed.

I believe this will pave the way better for the possible revenue share mode using Payment Splitter.

### DIFF
--- a/packages/contracts/contracts/deploy/PayableDAppFactory.sol
+++ b/packages/contracts/contracts/deploy/PayableDAppFactory.sol
@@ -27,21 +27,27 @@ contract PayableDAppFactory is
     Ownable,
     Pausable
 {
+    uint256 public immutable price;
+
     ENS public immutable ens;
     ICartesiDAppFactory public immutable factory;
     IERC20 public immutable token;
     IConsensus public immutable consensus;
-    uint256 public immutable price;
+    address public immutable payee;
 
     mapping(ICartesiDApp => uint256) public runway;
 
-    /// @notice Construct a new ERC20DAppFactory with the given token as payment method
+    /// @notice Construct a new PayableDAppFactory with the given token as payment method and price
+    /// @param _owner The owner of the factory, who can pause its operation
+    /// @param _payee The receiver of the payments to applications managed by this factory
     /// @param _factory The factory to delegate calls to instantiate the dapp
+    /// @param _token The token to use as payment method
     /// @param _consensus The consensus contract to use
     /// @param _price The price per second of execution of the dapp node
     constructor(
         address _owner,
         ENS _ens,
+        address _payee,
         ICartesiDAppFactory _factory,
         IERC20 _token,
         IConsensus _consensus,
@@ -52,6 +58,7 @@ contract PayableDAppFactory is
         token = _token;
         consensus = _consensus;
         price = _price;
+        payee = _payee;
 
         // transfer ownership to the give owner
         transferOwnership(_owner);
@@ -96,7 +103,7 @@ contract PayableDAppFactory is
         uint256 _runway
     ) external override whenNotPaused {
         require(
-            _dapp.getConsensus() == this.consensus(),
+            _dapp.getConsensus() == consensus,
             "PaymentDAppFactory: cannot import dapp with a different consensus"
         );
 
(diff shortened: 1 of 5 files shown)


felipeargento (dismissed):
Looks good!

But I would not make payee immutable and would add a new function to change it, in case one wants to swap their payment splitter contract.

Also, small note:
When you're declaring state variables, the order matters for gas-saving purposes.
Its cheaper to declare:
```
uint256
address
address
```

than
```
address
uint256
address
```

Storage is automatically packed, so it's better to order by type starting by the ones that use the full word (256bits) and then followed by the ones that can be packed together

tuler:
The problem is who is allowed to change the payee.
If the payee is a PaymentSplitter set by the PayableDAppSystem, the owner of PayableDAppFactory should not be allowed to change the payee (the PaymentSplitter).

apps/cli/src/deploy.ts
```diff
@@ -441,8 +441,9 @@ const deploy = async (
                     return true;
                 },
             });
+            const payee = await factory.read.payee();
             const { request } = await token.simulate.approve(
-                [factory.address, newAllowance],
+                [payee, newAllowance],
```
guidanoli, line 446: The tokens might be going to the payee now, but they are still managed by the factory, right?
As a result, this code section shouldn't need to change at all.

From [EIP-20](https://eips.ethereum.org/EIPS/eip-20):

> The `transferFrom` method is used for a withdraw workflow, allowing contracts to transfer tokens on your behalf. [...] The function SHOULD `throw` unless the `_from` account has deliberately authorized the sender of the message via some mechanism.

>

packages/contracts/contracts/deploy/PayableDAppFactory.sol
```diff
@@ -133,16 +140,16 @@ contract PayableDAppFactory is
...
 
         uint256 currentRunway = runway[_dapp];
 
         // if runway was in the past, update to now
-        currentRunway = currentRunway < block.timestamp
-            ? block.timestamp
-            : currentRunway;
+        if (currentRunway < block.timestamp) {
+            currentRunway = block.timestamp;
+        }
```
guidanoli, line 152: I don't understand why are you updating `currentRunway` if `< block.timestamp`.
Couldn't you call `extendRunway(dapp, 0)`, which would update the `currentRunway` to `block.timestamp`, and then later call `extendRunway` again with some actual value, saving you potentially a lot of tokens?

tuler: Suppose I pay for 1 week.
After 1 week the app stops.

1 month later I extend for 1 week.
I'm not paying for the 3 weeks it was stopped.
It's 1 week from now.

apps/cli/src/deploy.ts
guidanoli: Exactly, the recipient is not relevant in this case.
You can see how this specification is implemented by OpenZeppelin [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/a5ed318634016a25be4000ee07044a31f363e60c/contracts/token/ERC20/ERC20.sol#L163-L168).

packages/contracts/contracts/deploy/PayableDAppFactory.sol
guidanoli: Ohh, now I get it now. Yeah, that makes sense.

guidanoli (approved):

tuler:
@guidanoli how do you feel about @felipeargento ’s concern about payee immutability?

guidanoli:
> @guidanoli how do you feel about @felipeargento ’s concern about payee immutability?

I don't know how I feel about giving this power to the owner of the `PayableDAppFactory` contract. By default, I prefer contracts to be as trustless as possible. If you want to change the payee, deploy a new `PayableDAppFactory` contract via the `PayableDAppSystem` contract. Want to assign a completely different contract as the payee? Deploy your own flavor of `PayableDAppSystem` that allows an arbitrary pee.

This last bit could even be a feature: allow the user to decide whether they want to deploy a `PayableDAppFactory` with a `PaymentSplitter` contract as payee, or an arbitrary address instead. These could be two distinct functions.

guidanoli (approved):

PR #165, opened 2023-08-30
Feature/no backend

### LINKED ISSUE #158: Sunodo run with no backend
## 📚 Context

The rollups-examples includes a way to run Cartesi applications called [host mode](https://github.com/cartesi/rollups-examples#host-mode). The name is a bit confusing, but it means that the application code run on the host instead of running inside a Cartesi machine.

This is useful to speed up the development process, so the developer doesn't need to build a Cartesi machine and then execute a node.

There are trade-offs in this approach, like hiding the intrincasies of building software for RISC-V or allowing the application to execute outside the Cartesi machine sandbox and do a remote call to the internet. However the prototyping team still feels it's worth the risk.

This feature is implemented in the rollups-examples by replacing the server-manager container by another container using the image `cartesi/rollups-host-runner:1.0.0` instead of the `cartesi/server-manager:0.8.2`.

The `cartesi/rollups-host-runner:1.0.0` does not use the Cartesi machine at all.

## ✔️ Solution

Implement a `sunodo run --no-backend` option that will run the infrastructure of a rollups node that uses the `rollups-host-runner` instead of the `server-manager`.

"No backend" means that the node does not include the application backend. We believe this is a better term than using `host-mode`.

After the rollups-node is up the developer must run his application on his own, that will communicate with the node. He won't use `sunodo` to execute his application.

## 📈 Subtasks

- [ ] Build a rollups-node for this option. This can be a completely different image, or an option of the existing image.
- [ ] Add the `--no-backend` option to the `sunodo run` command

## 🎯 Definition of Done

-   [ ] `sunodo run --no-backend` working

oap75:
/close #158 

- [ ] add cartesi-host-runner to the rollups-node docker image
- [ ] add docker-compose for the host-runner
- [ ] add `--no-backend`


apps/cli/src/commands/run.ts
```diff
@@ -20,6 +20,10 @@ export default class Run extends Command {
             description: "duration of an epoch (in seconds)",
             default: 86400,
         }),
+        "no-backend": Flags.boolean({
+            description: "disable backend and snapshot check",
```
tuler, line 24: Set the `summary` field to `run a node without the application code.`
In the `description`, which is a long description use:

```
Run a node without the application code. Application must be executed by the developer on the host machine, fetching inputs from the node running at http://localhost:5004
```

Or whatever port that will be.

packages/rollups-node/Dockerfile
```diff
@@ -8,6 +8,7 @@ FROM ${REGISTRY}/${ORG}/rollups-dispatcher:${ROLLUPS_VERSION}      AS dispatcher
 FROM ${REGISTRY}/${ORG}/rollups-advance-runner:${ROLLUPS_VERSION}  AS advance_runner
 FROM ${REGISTRY}/${ORG}/rollups-inspect-server:${ROLLUPS_VERSION}  AS inspect_server
 FROM ${REGISTRY}/${ORG}/rollups-graphql-server:${ROLLUPS_VERSION}  AS graphql_server
+FROM ${REGISTRY}/${ORG}/rollups-host-runner:1.0.0                  AS host-runner
```
endersonmaia, line 11:
```suggestion
FROM ${REGISTRY}/${ORG}/rollups-host-runner:${ROLLUPS_VERSION}     AS host-runner
```

apps/cli/src/commands/run.ts
```diff
@@ -31,18 +35,23 @@ export default class Run extends Command {
...
+                    "Cartesi machine snapshot not found, run 'sunodo build'"
+                );
+            }
+            // read hash of the cartesi machine snapshot
+            const hash = fs
+                .readFileSync(path.join(snapshot, "hash"))
+                .toString("hex");
+            projectName = hash.substring(0, 8);
+        } else {
+            projectName = "localhost"; // Set project name to "localhost" when use --no-backend flag
```
tuler, line 52: Use `sunodo-node` as the project name.

packages/rollups-node/s6-overlay/s6-rc.d/host-runner/run
tuler, line 1: The host-runner must only run in case of `no-backend`

endersonmaia (changes requested):
Can you guide me on how I can test this?

packages/rollups-node/s6-overlay/s6-rc.d/host-runner/run
endersonmaia: An environment variable could be used as a flag, look at https://github.com/sunodo/sunodo/blob/main/packages/rollups-node/s6-overlay/s6-rc.d/stream-cleanup/script/up.sh#L3 where I flag the stream cleanup oneshot script.

tuler:
The rollups-host-runner runs by default on port 5004.
```
docker run cartesi/rollups-host-runner:1.0.0 --help
```

This is the port the application running on the host will have to communicate with.
We should probably use the node `traefik` to route to that service, the same way we are doing for the inspect and graphql server, and only occupy the 8080 port of the host.

packages/rollups-node/Dockerfile
oap75: The `rollups-host-runner` docker image  versions is not as same as other services : 

https://hub.docker.com/r/cartesi/rollups-host-runner/tags

endersonmaia: rollups-host-runner` was just released at 1.0.0, previously it was on a different repository and name https://github.com/cartesi/host-server-manager
 
 reusing `ROLLUPS_VERSION` for rollups-host-runner is safe here, we could even consider renaming it to `ROLLUPS_NODE_VERSION`, to be more precise

endersonmaia (changes requested):
I still need to execute and test, will come back with more feedback after that.

About using the `entrypoint.sh` to change the `s6`, I think it's not the place, let's try to use `s6` features to control which services should run or not.

apps/cli/src/commands/run.ts
```diff
@@ -87,14 +106,18 @@ export default class Run extends Command {
 
         try {
             // run compose environment
-            await execa("docker", [...args, "up", ...attachment], {
+            const upCommand = `docker-compose ${args.join(" ")} up ${attachment.join(" ")}`;
```
endersonmaia, line 109:
```suggestion
            const upCommand = `docker ${args.join(" ")} up ${attachment.join(" ")}`;
```

apps/cli/src/commands/run.ts
```diff
@@ -87,14 +106,18 @@ export default class Run extends Command {
...
+            await execa("sh", ["-c", upCommand], {
                 env,
                 stdio: "inherit",
+                shell: true,
             });
         } finally {
             // shut it down, including volumes
-            await execa("docker", [...args, "down", "--volumes"], {
+            const downCommand = `docker-compose ${args.join(" ")} down --volumes`;
```
endersonmaia, line 117:
```suggestion
            const downCommand = `docker ${args.join(" ")} down --volumes`;
```

endersonmaia (changes requested):

apps/cli/src/commands/run.ts
```diff
@@ -63,18 +73,27 @@ export default class Run extends Command {
...
+        );
+
+        // Use composeFileDev by default, add composeFileHost if --no-backend is defined
+        const selectedComposeFile = flags["no-backend"] ? [composeFileDev, composeFileHost] : composeFileDev;
 
         const args = [
             "compose",
             "--file",
-            composeFile,
+            selectedComposeFile,
```
endersonmaia, line 96: we need a new `--file` for each file, like:

```
docker compose --file docker-compose-dev.yml --file docker-compose-host.yml ...
```

endersonmaia (changes requested):

apps/cli/src/commands/run.ts
tuler: Actually the documentation says a CSV is accepted.

But I think I prefer a profile definition

apps/cli/src/commands/run.ts
```diff
@@ -87,14 +106,18 @@ export default class Run extends Command {
 
         try {
             // run compose environment
-            await execa("docker", [...args, "up", ...attachment], {
+            const upCommand = `docker ${args.join(" ")} up ${attachment.join(" ")}`;
```
tuler, line 109: I prefer not to do `join`, pass the list of arguments to `execa`

tuler:
> About using the `entrypoint.sh` to change the `s6`, I think it's not the place, let's try to use `s6` features to control which services should run or not.

I agree. What is the "s6 way" of conditionally executing a service?

apps/cli/src/commands/run.ts
endersonmaia: CSV didn't worked for me :thinking:

Agree that profile is a cleaner solution.

oap75: > But I think I prefer a profile definition

I think we can't use profile, because the only thing we need is to replace some ENVs and I think having an override compose file is better solution.

oap75:
> > > About using the `entrypoint.sh` to change the `s6`, I think it's not the place, let's try to use `s6` features to control which services should run or not.
> > 
> > 
> > I agree. What is the "s6 way" of conditionally executing a service?
> 
> We could use this `S6_STAGE2_HOOK` environment variable.

Fixed based on `S6_STAGE2_HOOK`

packages/rollups-node/traefik/conf.d/host-runner.yaml
```diff
@@ -0,0 +1,10 @@
+http:
+  routers:
+    inspect_server:
```
endersonmaia, line 3:
```suggestion
    host_runner:
```

packages/rollups-node/traefik/conf.d/host-runner.yaml
```diff
@@ -0,0 +1,10 @@
...
+      rule: "PathPrefix(`/`)"
+      service: host-runner
+  services:
+    inspect_server:
```
endersonmaia, line 7:
```suggestion
    host_runner:
```

packages/rollups-node/stage2-hook.sh
endersonmaia, line 1: I'd put this file inside `packages/rollups-node/s6-overlays/`

endersonmaia (changes requested):

packages/rollups-node/Dockerfile
oap75: Fixed !

endersonmaia (changes requested):
I basically reviewed the commits' content.

two commit may be enough, first for the `rollups-node` second one for the `cli`.

Remember to add the `changesets` for each one.

apps/cli/src/node/docker-compose-dev.yml
```diff
@@ -92,6 +92,7 @@ services:
             - 8080:8080
         environment:
             # s6-overlay
+            S6_STAGE2_HOOK: "/etc/s6-overlay/scripts/stage2-hook.sh"
```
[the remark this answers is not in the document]
oap75: We need it into the docker-compose-dev because of the script content : 

```
else
    /command/s6-echo "host-runner: ENABLE_SUNODO_NODE is not set, skip the host mode..."
    sed -i '/host-runner/d' /usr/local/bin/is_ready
    rm /etc/traefik/conf.d/host-runner.yaml
fi
```

packages/rollups-node/Dockerfile
```diff
@@ -64,14 +65,17 @@ COPY --from=dispatcher      /opt/cartesi/bin/cartesi-rollups-dispatcher     /usr
...
 RUN mkdir -p /var/opt/cartesi/machine-snapshots/0_0 && ln -s /var/opt/cartesi/machine-snapshots/0_0 /var/opt/cartesi/machine-snapshots/latest
 
 COPY ./traefik /etc/traefik
-COPY ./s6-overlay /etc/s6-overlay
+COPY ./s6-overlay/s6-rc.d /etc/s6-overlay/s6-rc.d
+COPY ./s6-overlay/stage2-hook.sh /etc/s6-overlay/scripts/stage2-hook.sh
 COPY ./is_ready /usr/local/bin/
 COPY ./address_book /usr/local/bin/
 COPY ./entrypoint.sh /entrypoint.sh
+COPY ./stage2-hook.sh /etc/cont-init.d/
```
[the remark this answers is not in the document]
oap75: Fixed !

apps/cli/src/node/docker-compose-host.yml
```diff
@@ -0,0 +1,27 @@
...
+                "--outputFile",
+                "/usr/share/sunodo/dapp.json",
+            ]
+
+    validator:
+        environment:
+            ENABLE_SUNODO_NODE: true
+
+            SERVER_MANAGER_ENDPOINT: http://localhost:5004
+            SERVER_MANAGER_ADDRESS: localhost:5004
```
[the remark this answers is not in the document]
oap75: Fixed !

packages/rollups-node/Dockerfile
```diff
@@ -8,6 +8,7 @@ FROM ${REGISTRY}/${ORG}/rollups-dispatcher:${ROLLUPS_VERSION}      AS dispatcher
 FROM ${REGISTRY}/${ORG}/rollups-advance-runner:${ROLLUPS_VERSION}  AS advance_runner
 FROM ${REGISTRY}/${ORG}/rollups-inspect-server:${ROLLUPS_VERSION}  AS inspect_server
 FROM ${REGISTRY}/${ORG}/rollups-graphql-server:${ROLLUPS_VERSION}  AS graphql_server
+FROM ${REGISTRY}/${ORG}/rollups-host-runner:${ROLLUPS_VERSION}     AS host-runner
```
endersonmaia, line 11:
```suggestion
FROM ${REGISTRY}/${ORG}/rollups-host-runner:${ROLLUPS_VERSION}     AS host_runner
```

packages/rollups-node/Dockerfile
```diff
@@ -64,11 +65,13 @@ COPY --from=dispatcher      /opt/cartesi/bin/cartesi-rollups-dispatcher     /usr
 COPY --from=advance_runner  /opt/cartesi/bin/cartesi-rollups-advance-runner /usr/bin/cartesi-rollups-advance-runner
 COPY --from=inspect_server  /opt/cartesi/bin/cartesi-rollups-inspect-server /usr/bin/cartesi-rollups-inspect-server
 COPY --from=graphql_server  /opt/cartesi/bin/cartesi-rollups-graphql-server /usr/bin/cartesi-rollups-graphql-server
+COPY --from=host-runner     /opt/cartesi/bin/cartesi-rollups-host-runner    /usr/bin/cartesi-rollups-host-runner
```
endersonmaia, line 68:
```suggestion
COPY --from=host_runner     /opt/cartesi/bin/cartesi-rollups-host-runner    /usr/bin/cartesi-rollups-host-runner
```

endersonmaia (changes requested):

packages/rollups-node/Dockerfile
oap75: Fixed !

packages/rollups-node/Dockerfile
oap75: Fixed !

endersonmaia (approved):

PR #180, opened 2023-09-01
Fetch entire git repo, including tags

tuler:
Fact: we need to fetch tags on the release workflow.

The code below only fetches tags which belong to the fetched history, which in this case is only 1 commit, as the default `fetch-depth` is 1.

```
uses: actions/checkout@v3
with:
    fetch-tags: true
```

So we will need to checkout the entire history by using `fetch-depth` zero, which includes all tags.
```
uses: actions/checkout@v3
with:
    fetch-depth: 0
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,7 +21,7 @@ jobs:
             - name: Checkout Repo
               uses: actions/checkout@v3
               with:
-                  fetch-tags: true
+                  fetch-depth: 0
 
             - name: Setup Node.js
               uses: actions/setup-node@v3


tuler:
It works now!

```
$ /home/runner/work/sunodo/sunodo/node_modules/.bin/changeset tag
🦋  Skipping tag (already exists):  @sunodo/cli@0.7.0
🦋  Skipping tag (already exists):  docs@0.0.0
🦋  Skipping tag (already exists):  web@0.0.0
🦋  Skipping tag (already exists):  anvil@3.0.0
🦋  Skipping tag (already exists):  @sunodo/contracts@0.2.0
🦋  Skipping tag (already exists):  @sunodo/devnet@1.1.0
🦋  Skipping tag (already exists):  eslint-config-custom@0.0.0
🦋  Skipping tag (already exists):  rollups-node@0.3.0
🦋  Skipping tag (already exists):  @sunodo/sdk@0.2.0
🦋  Skipping tag (already exists):  @sunodo/token@0.2.0
🦋  Skipping tag (already exists):  tsconfig@0.0.0
🦋  Skipping tag (already exists):  ui@0.0.0
Everything up-to-date
```

PR #183, opened 2023-09-01
Change default template repo branch

tuler:
Changed default branch of sunodo/sunodo-templates to `sdk-0.2`.
Already created the branch `sdk-0.2` in that repository.

### DIFF
--- a/.changeset/proud-foxes-fetch.md
+++ b/.changeset/proud-foxes-fetch.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/cli": patch
+---
+
+change default sunodo/sunodo-templates repository default branch the create command uses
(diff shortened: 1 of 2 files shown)


endersonmaia (approved):
Since we know that a branch can update which `ref` it's pointing to during time, I think it's important to log the commit hash that the `sunodo create` command is using.

```
❯ ./bin/run.js create --template javascript my-js-dapp
✔ Creating from sunodo-templates/javascript (sdk-0.2@7b0a7150675ae7229ac5a556926ee78e18a7136c)
✔ Application created at /home/endersonmaia/Development/sunodo/sunodo/apps/cli/my-js-dapp
```

Something like that would help during troubleshooting problems reported by users.

tuler:
> Since we know that a branch can update which `ref` it's pointing to during time, I think it's important to log the commit hash that the `sunodo create` command is using.

I agree. Can you open another issue for that?

endersonmaia:
> > Since we know that a branch can update which `ref` it's pointing to during time, I think it's important to log the commit hash that the `sunodo create` command is using.
> 
> I agree. Can you open another issue for that?

https://github.com/sunodo/sunodo/issues/187

PR #190, opened 2023-09-05
feat(cli): show commit hash of created template

### LINKED ISSUE #187: sunodo create should log commit hash reference for the template used
## 📚 Context

When using  `sunodo create`, templates available at the sunodo/sunodo-templates repository will be used after a branch defined by the sunodo/sdk version, for ex.: `sdk-0.2`.

A branch can change which Git RE  it's pointing to over time.

The user can also define another branch using the `--branch` argument, and that should override the branch used.

So,ion case a user is troubleshooting problems with applications created using a `sunodo create`, it's important to have the specific Git REF.

## ✔️ Solution

Print the Git REF used for that `sunodo create` call.

For ex.:

```
❯ ./bin/run.js create --template javascript my-js-dapp
✔ Creating from sunodo-templates/javascript (sdk-0.2@7b0a7150675ae7229ac5a556926ee78e18a7136c)
✔ Application created at /home/endersonmaia/Development/sunodo/sunodo/apps/cli/my-js-dapp
```

endersonmaia:
/close #187

### DIFF
--- a/apps/cli/src/commands/create.ts
+++ b/apps/cli/src/commands/create.ts
@@ -65,11 +65,13 @@ export default class CreateCommand extends Command {
         const { args, flags } = await this.parse(CreateCommand);
         const spinner = ora("Creating application...").start();
         try {
-            const { dir } = await this.download(
+            const { dir, commit } = await this.download(
                 flags.template,
                 flags.branch,
                 args.name,
             );
+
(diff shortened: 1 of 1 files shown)


endersonmaia:
Since the lib we're using downloads the archive via HTTP and do not use any git protocol, there's no way to get this information directly.

I opened an [issue upstream](https://github.com/unjs/giget/issues/106) to implement the `git get-tar-commit-id` equivalent to giget.

PR #194, opened 2023-09-07
Fix address-book so getting dapp address works with no-backend

### LINKED ISSUE #154: Sunodo send must ask for DApp address
## 🙂 Expected behavior

`sunodo send` must ask for DApp address if not defined by the command flag `--dapp`.

## 🫠 Actual behavior

If not defined by `--dapp` the address is being resolved from the local execution environment address of deployed DApp.

But the destination of input can be a testnet or mainnet environment, so it doesn't make sense.

## 🧪 Minimal test case

Run `sunodo send` and choose a testnet.

## ✔️ Possible solutions

We can still resolve the local execution DApp address as the default input value in case the chosen chain is the devnet.

tuler:
1) simplification of rollups contracts of address-book, using values from the `contracts.ts` file that are already imported into the project by a build procedure.

2) the dapp address is read directly from a running anvil container, either of a normal node, or a no-backend node. I use cat to read it, instead of saving a host file with the address.

3) the send command now always asks for the address, and the one from (2) is presented as default value only.

`sunodo address-book` should print even without any node running, in that case without CartesiDApp address
`sunodo address-book` should also print dapp address if there is a running node, backend or no-backend

closes #154

### DIFF
--- a/.changeset/young-kids-relate.md
+++ b/.changeset/young-kids-relate.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/cli": patch
+---
+
+fix address-book so getting dapp address works with no-backend
--- a/apps/cli/src/commands/send/index.ts
+++ b/apps/cli/src/commands/send/index.ts
@@ -70,14 +70,15 @@ export abstract class SendBaseCommand<
             return this.flags.dapp;
         }
 
-        // get the address book
-        const addressBook = await super.getAddressBook();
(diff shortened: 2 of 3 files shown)


endersonmaia (approved):
`sunodo send` works with a node running with the caveat that you have to run it from inside the dapp directory you did run the `sunodo run`.

PR #197, opened 2023-09-08
feat(docs): no-backend mode

tuler:
Basic documentation about `run --no-backend` mode.

### DIFF
--- a/apps/docs/guide/running/running-application.md
+++ b/apps/docs/guide/running/running-application.md
@@ -1,3 +1,30 @@
 # Running application
 
 The `sunodo run` command will execute a Cartesi node for the application previously built with `sunodo build`.
+
+## Verbosity
+
+By default the node works on non-verbose mode, and only outputs logs coming from the user backend application. In case the user needs more information there is the `--verbose` command option.
+
+## Blockchain configuration
+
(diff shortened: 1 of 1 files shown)


apps/docs/guide/running/running-application.md
```diff
@@ -1,3 +1,30 @@
...
+## No-backend mode
+
+Sunodo `run` also supports running a node without the user application backend packaged as a Cartesi machine. In this case the user application can be executed on the host, without the concern of being compiled to RISC-V.
+
+This is a useful development workflow help, but there are some caveats the user must be aware:
+
+1. the application will eventually need to be compiled to RISC-V, or use a RISC-V runtime in case of interpreted languages;
+2. in this mode the application will not be running inside the sandbox of the Cartesi machine, and will not block operations that won't be allowed when running inside a Cartesi machine, like accessing remote resources;
+3. this mode only works for applications that use the Cartesi Rollups HTTP API, and doesn't work with applications using the low-level API;
+4. Performance inside a Cartesi machine will be much lower than running on the host.
```
endersonmaia, line 28:
```suggestion
1. The application will eventually need to be compiled to RISC-V, or use a RISC-V runtime in case of interpreted languages;
2. In this mode the application will not be running inside the sandbox of the Cartesi machine, and will not block operations that won't be allowed when running inside a Cartesi machine, like accessing remote resources;
3. This mode only works for applications that use the Cartesi Rollups HTTP API, and doesn't work with applications using the low-level API;
4. Performance inside a Cartesi machine will be much lower than running on the host.
```

endersonmaia (changes requested):

endersonmaia (approved):

PR #199, opened 2023-09-08
feat(cli): support .sunodo.env for sunodo run

endersonmaia:
This PR will make `sunodo run` accept a `.sunodo.env` file which make the sunodo/rollups-node container load the environment variables, so that one can pass variables to the sunodo/rollups-node services.

### DIFF
--- a/apps/cli/src/commands/run.ts
+++ b/apps/cli/src/commands/run.ts
@@ -77,43 +77,31 @@ export default class Run extends Command {
             WS_URL: "ws://anvil:8545",
         };
 
-        // Construct the path for Docker Compose files
-        const composeFileDev = path.join(
+        const composeFiles = [];
+        const composeBasePath = path.join(
             "--file=",
             path.dirname(new URL(import.meta.url).pathname),
             "..",
-            "node",
-            "docker-compose-dev.yml",
-        );
-        const composeFileHost = path.join(
-            "--file=",
-            path.dirname(new URL(import.meta.url).pathname),
-            "..",
-            "node",
-            "docker-compose-host.yml",
-        );
+            "node"
+        )
 
-        // Use composeFileDev by default, add composeFileHost if --no-backend is defined
-        const selectedComposeFiles = flags["no-backend"]
-            ? [composeFileDev, composeFileHost]
-            : [composeFileDev];
+        const composeFileDev = path.join(composeBasePath, "docker-compose-dev.yml");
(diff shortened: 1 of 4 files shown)


apps/cli/src/commands/run.ts
```diff
@@ -108,6 +108,11 @@ export default class Run extends Command {
         ];
         const attachment = flags.verbose ? [] : ["--attach", "validator"];
 
+        // create en empty .sunodo.env file if it doesn't exists
+        if (!fs.existsSync(".sunodo.env")) {
+            fs.writeFileSync(".sunodo.env", "");
```
tuler, line 113: research a way to not need this

apps/docs/guide/running/running-application.md
```diff
@@ -16,6 +16,21 @@ The private chain by default has a block time of 5 seconds, controlled by `--blo
...
+All Rollups Node services can be configured using environment variables.
+
+For ex., in case you want to change the default deadline for adanving the state for the rollups-advance-runner service.
+
+```shell
+cat .sunodo.env
+SM_DEADLINE_ADVANCE_STATE=360000
+```
+
+Check the configuration options for each service in the [Rollups Node documentation](https://github.com/cartesi/rollups/blob/v1.0.1/offchain/README.md).
```
tuler, line 32: ps: except that it's not documented there

apps/cli/src/commands/run.ts
endersonmaia: already did, `compose` fails if the file doesn't exists

what I can do is create a `docker-compose-envfile.yaml` that would be added to `compose --file` only if the file exists.

apps/docs/guide/running/running-application.md
endersonmaia: yeah, what do you suggest?

I thought on instructing the user to nu `docker run ... cartesi/rollups-advance-runner --help` but I felt it's out of scope for this documentation, and ugly.

apps/cli/src/node/docker-compose-dev.yml
```diff
@@ -90,6 +90,8 @@ services:
             retries: 5
         ports:
             - 8080:8080
+        env_file:
+            - ./.sunodo.env
```
tuler, line 94: Does this should be here? I see you are adding a compose file if the env file exists.

endersonmaia: Fixed!

tuler (approved):

PR #219, opened 2023-09-27
Explorer integration

tuler:
- moves traefik from rollups-node to the compose, so we can have more control to add other services to the compose environment and maintain a single entry port
- add the rollups explorer to the compose


apps/cli/src/node/docker-compose-dev.yml
```diff
@@ -76,7 +76,7 @@ services:
             - ./.sunodo/image:/var/opt/cartesi/machine-snapshots/0_0:ro
 
     validator:
-        image: sunodo/rollups-node:0.4.1
+        image: sunodo/rollups-node:devel
```
endersonmaia, line 79: Can we use some kind of variable here? 

To differ when calling from a `sunodo-dev` alias and from a `sunodo` cli installed.

apps/cli/src/commands/run.ts
```diff
@@ -86,7 +86,10 @@ export default class Run extends Command {
 
         // construct the path for Docker Compose files
         const composePath = (file: string) => path.join(binPath, "node", file);
-        const composeFiles = ["docker-compose-dev.yml"];
+        const composeFiles = [
+            "docker-compose-dev.yml",
+            "docker-compose-explorer.yml",
```
endersonmaia, line 91: Won't explorer be optional?

apps/cli/src/node/docker-compose-explorer.yml
```diff
@@ -0,0 +1,44 @@
...
+            PGPORT: 5432
+            PGUSER: postgres
+            PGPASSWORD: password
+            PGDATABASE: postgres
+        depends_on:
+            database:
+                condition: service_healthy
+
+    squid_api:
+        image: cartesi/rollups-explorer-api:0.1.0-rc.0
```
endersonmaia, line 18: IF we can have a variable here, it will ake it easy when updating.

apps/cli/src/node/docker-compose-explorer.yml
```diff
@@ -0,0 +1,44 @@
...
+            GQL_PORT: 4350
+        ports:
+            - "4350:4350"
+        command: ["sqd", "serve:prod"]
+        depends_on:
+            database_creator:
+                condition: service_completed_successfully
+
+    squid_processor:
+        image: cartesi/rollups-explorer-api:0.1.0-rc.0
```
endersonmaia, line 33: IF we can have a variable here, it will ake it easy when updating.

endersonmaia (commented):
Some comment, I know it's still a draft ;)

apps/cli/src/commands/run.ts
tuler: What is your suggestion about how to enable/disable various services of the local env?

apps/cli/src/node/docker-compose-explorer.yml
tuler: I don't think this is something we have to allow sunodo users to control.
Sunodo developers can just edit this file on their local dev environment.

apps/cli/src/commands/run.ts
endersonmaia: We could split the `services:` definition in different `.yaml` files, and use some flag like `--disable-explorer`, composing the `composeFiles` array.

We need to remember that the order of compose files is important.

tuler: I think having `--enable-<service>` and `--disable-<service>` flags will increase a lot the number of flags as we add services.

What about `--services=<service1>,<service2>`

endersonmaia: What services are we talking about? The ones defined at compose file, right? And not the ones inside the rollups-node.

Considering all services are enabled by default, we could have a `--disable-services=[...]`

endersonmaia: Put that in an Issue, please.

tuler: Will do when the idea solidifies. For this issue I’ll code as including it by default.

apps/cli/src/node/docker-compose-explorer.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+    squid_processor:
+        image: cartesi/rollups-explorer-api:0.1.0
+        environment:
+            CHAIN_ID: 31337
+            RPC_ENDPOINT: http://anvil:8545
+            DB_NAME: squid
+            DB_PORT: 5432
+            DB_HOST: database
+            DB_PASS: password
```
endersonmaia, line 40: Could use YAML anchor tag to avoid duplicating this configuration.

endersonmaia (changes requested):
Minor comment about YAML anchor, LGTM, remove `DELETE:` and it's goog to go.

tuler:
I think now we can consider if this will be included in 0.9, or if 0.9 will be only rollups-node 0.1.0

apps/cli/src/node/docker-compose-explorer.yaml
tuler: I don't think it's a big deal for now

endersonmaia (approved):

PR #224, opened 2023-09-29
Remove traefik from node

tuler:
This PR removes traefik from inside the rollups-node and exposes the rollups-node ports to the internal docker network, so a traefik in that network can route to the services (graphql, inspect, host-runner).


endersonmaia (changes requested):
You forgot to remove the port mapping from the `docker-compose-dev.yaml`

https://github.com/sunodo/sunodo/blob/feature/remove-node-traefik/apps/cli/src/node/docker-compose-dev.yml#L85

endersonmaia (approved):
I saw you did on the other PR.

tuler:
> You forgot to remove the port mapping from the `docker-compose-dev.yaml`
> 
> https://github.com/sunodo/sunodo/blob/feature/remove-node-traefik/apps/cli/src/node/docker-compose-dev.yml#L85

The idea of this PR is to not touch the CLI, so we can first release a new version of the node, and then release a version of the CLI that uses that node.

PR #231, opened 2023-10-03
Prompt of running services

tuler:
This is how it looks:

```
79c354a7-validator-1  | host-runner: ENABLE_SUNODO_NODE is not set, skip the host mode...
79c354a7-validator-1  | Provisioning cartesi-machine snapshot...
79c354a7-prompt-1     | Anvil running at http://localhost:8545
79c354a7-prompt-1     | GraphQL running at http://localhost:8080/graphql
79c354a7-prompt-1     | Inspect running at http://localhost:8080/inspect/
79c354a7-prompt-1     | Press Ctrl+C to stop the node
79c354a7-prompt-1 exited with code 0
```

I'm not super happy with the last line, because it may give the user the impression something went wrong.
Any suggestion?

### DIFF
--- a/apps/cli/src/node/docker-compose-dev.yaml
+++ b/apps/cli/src/node/docker-compose-dev.yaml
@@ -160,11 +160,30 @@ services:
         depends_on:
             validator:
                 condition: service_healthy
+        healthcheck:
+            test: ["CMD", "traefik", "healthcheck"]
+            interval: 5s
+            timeout: 1s
+            retries: 5
         ports:
             - 8080:8080
-            - 8088:8088
         volumes:
             - ${SUNODO_BIN_PATH}/node/traefik:/etc/traefik
+
+    prompt:
+        image: debian:bookworm-slim
+        command:
+            [
+                "sh",
+                "-c",
+                "ls | sort | xargs cat; trap exit INT TERM; sleep infinity & wait",
+            ]
+        working_dir: /prompt
+        volumes:
+            - ${SUNODO_BIN_PATH}/node/prompt:/prompt
+        depends_on:
+            proxy:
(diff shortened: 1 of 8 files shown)


endersonmaia:
What if you use `sleep infinity` as a command inside a bash entrypoint that prints the medsage instead of this multiple files approach?

tuler:
If I add a `; sleep infinity` to the container command, the container is not being gracefully shutdown.

tuler:
It's working now.

I feed it adds a few seconds to the startup process, as we have more healthy waits.

apps/cli/src/node/docker-compose-dev.yaml
```diff
@@ -160,11 +160,30 @@ services:
...
+            timeout: 1s
+            retries: 5
         ports:
             - 8080:8080
-            - 8088:8088
         volumes:
             - ${SUNODO_BIN_PATH}/node/traefik:/etc/traefik
 
+    prompt:
+        image: debian:bookworm-20230814
```
endersonmaia, line 174: the user should already have `debian:bookworm-slim` because of `cartesi/rollups-node`.

https://github.com/cartesi/rollups-node/blob/main/build/Dockerfile#L18C7-L18C26

```suggestion
        image: debian:bookworm-slim
```

endersonmaia (changes requested):
Don't forget to squash.

tuler:
Please review

endersonmaia (approved):
About the time to be available, it takes 30s on my machine, don't know how to reduce it.

I tried some changes on `interval` and `timeout` but it didn't have any impact.

tuler:
Yes, I think it's too long, but I also don't understand well why.

PR #233, opened 2023-10-03
Refactoring contracts to introduce node providers

tuler:
This PR implements the model below.
image

PR #234, opened 2023-10-03
Indexing new providers contracts

tuler:
Squid implementation of indexing of application nodes.

We can also consider to merge this with the indexing done for the [rollups-explorer](https://github.com/cartesi/rollups-explorer-api), because they have a lot in common. We can see this as a layer on top of that.

Check contracts at https://github.com/sunodo/sunodo/pull/233


tuler:
If we merge this to the schema of the cartesi rollups explorer index this is what it would look like:

image

PR #244, opened 2023-10-05
Update rollups-explorer images version

brunomenezes:
### Summary
Changes to include latest rollups-explorer-api image and also update the environment variable for RPC setup.

### DIFF
--- a/.changeset/gorgeous-starfishes-yell.md
+++ b/.changeset/gorgeous-starfishes-yell.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/cli": minor
+---
+
+Upgrade rollups-explorer docker images UI(v0.9.0), API(0.4.0)
--- a/apps/cli/src/node/docker-compose-explorer.yaml
+++ b/apps/cli/src/node/docker-compose-explorer.yaml
@@ -12,7 +12,7 @@ services:
             PGDATABASE: ${PGDATABASE:-postgres}
 
     explorer_api:
-        image: cartesi/rollups-explorer-api:0.1.0
+        image: cartesi/rollups-explorer-api:0.4.0
         environment:
             DB_NAME: ${SUNODO_SQUID_DB_NAME:-squid}
             DB_PORT: ${PGPORT:-5432}
@@ -27,10 +27,10 @@ services:
                 condition: service_completed_successfully
 
     squid_processor:
-        image: cartesi/rollups-explorer-api:0.1.0
+        image: cartesi/rollups-explorer-api:0.4.0
         environment:
             CHAIN_ID: ${CHAIN_ID:-31337}
-            RPC_ENDPOINT: ${RPC_URL:-http://anvil:8545}
+            RPC_URL_31337: ${RPC_URL:-http://anvil:8545}
             DB_NAME: ${SUNODO_SQUID_DB_NAME:-squid}
             DB_PORT: ${PGPORT:-5432}
             DB_HOST: ${PGHOST:-database}
@@ -41,7 +41,7 @@ services:
                 condition: service_completed_successfully
 
     explorer:
-        image: cartesi/rollups-explorer:0.1.1
+        image: cartesi/rollups-explorer:0.9.0
         expose:
             - 3000
         depends_on:


tuler:
`cartesi/rollups-explorer-api` is used twice in the compose file.

Also add a changeset.

I'd rather bump the version of explorer in sunodo when we have actual new features or bug fixes to include. Because we need to make a sunodo release for this to go live.

But this PR can stay open until we have that.

brunomenezes:
> `cartesi/rollups-explorer-api` is used twice in the compose file.
> 
> Also add a changeset.
> 
> I'd rather bump the version of explorer in sunodo when we have actual new features or bug fixes to include. Because we need to make a sunodo release for this to go live.
> 
> But this PR can stay open until we have that.

Cool, I changed the PR to draft. When the time is right we make the appropriate updates here.

tuler:
Can we move with this PR?

tuler:
The following route is giving me a 404
http://localhost:8080/explorer/applications/0x70ac08179605af2d9e75782b8decdd3c22aa4d0c

tuler:
I sent a raw input and it did not work correctly.
I sent `0x4020` as input, and what went into the application was 0x307834303230
If I send `0x4020` using `sunodo send` it works correctly.

brunomenezes:
I had a headwind with Sunodo. But here is what I found;

> The following route is giving me a 404 http://localhost:8080/explorer/applications/0x70ac08179605af2d9e75782b8decdd3c22aa4d0c

It is a network problem. That route `/applications/[address]` tries to reach out to an external resource (explorer_api) to check if the Application does exist, so it is an `egress` problem since it is trying to use the graphql_url configuration (i.e. `http://localhost:8080/explorer-api/graphql`) I notice the docker network, in that case, has a few aliased. Once I added `http://explorer_api:4350/graphql`, it was able to reach out to the server internally. Independent of aliases, it is necessary to add a separate endpoint with the correct **hostname** and for this, I will add a new env variable to cover that internal case. I am open to options but don't see another way for that case. cc: @endersonmaia 

> I tried to make a ERC-20 deposit using the SunodoToken that sunodo has (0xf795b3D15D47ac1c61BEf4Cc6469EBb2454C6a9b), but got the error below during screen validation.
> 
> The contract function "decimals" returned no data ("0x").

Are you sure about that address? Every time I stop and do a `sunodo run`, the SunodoToken address is (0xae7f61eCf06C65405560166b259C54031428A9C4) and is working as expected. So, the error message is working as expected.

image


>  sent a raw input and it did not work correctly.
I sent 0x4020 as input, and what went into the Application was 0x307834303230
If I send 0x4020 using sunodo send it works correctly.

That is not working as expected. We're checking if the input is a `hex` value and if not displaying some error messages (👍 ), but we are applying `toHex()` on top of the already hex value. We'll do a fix.

tuler:
You are probably right about the token address. I was getting the address from `sunodo address-book`, but I think that is wrong, using and old deployment.

tuler:
About the 404, do we need to do a server-side check?

endersonmaia:
> I will add a new env variable to cover that internal case

What if `explorer` and `explorer_api` stay in a single container?

brunomenezes:
> About the 404, do we need to do a server-side check?

for context is a server-side rendering action, not just a check; if the app does not exist, 404 is returned instead of returning the page, hydration kick-in, followed by queries to list the inputs of such app. We're just leveraging the NextJS 13+ and App router capabilities.

brunomenezes:
> > I will add a new env variable to cover that internal case
> 
> What if `explorer` and `explorer_api` stay in a single container?

I see, It would still need a different URL for it to communicate, like `http://localhost:4350/graphql`, which is different from the external one used by the client side (i.e. the App running in the Browser).

endersonmaia (dismissed):

brunomenezes:
I put the PR on draft again, because new features and fixes are coming:

* ERC-721 deposit support (Already live)
* Voucher Execution (On review / testing)

brunomenezes:
Moving the PR from draft to ready to review.
cc: @tuler , @endersonmaia

tuler (approved):

PR #250, opened 2023-10-09
Bump dependencies

### LINKED ISSUE #176: Brew uses a package with no lock file
## 🙂 Expected behavior

Brew should install the sunodo "binary" the same way as built locally in the monorepo project.

ps: sunodo is not a native binary, it's a javascript entrypoint.

## 🫠 Actual behavior

Our [brew formula](https://github.com/sunodo/homebrew-tap/blob/main/Formula/sunodo.rb) downloads the [npm package](https://registry.npmjs.org/@sunodo/cli/-/cli-0.7.0.tgz) from npmjs.
This npm package does not have a [package-lock.json](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json) file.

After downloading the npm package it runs `npm install`.
This triggers the npm version resolution strategy from the `package.json` file.
The `package.json` defines dependencies very loosely, on purpose, like `view ^1.9`, or `execa ^8.0`, expecting that the lock file would guarantee the reproducibility of the build.

## 🧪 Minimal test case

I don't have a test case.
The brew formula for CLI 0.7.0 is failing, but I'm still investigating and it can be for a different reason.

## 🌎 Environment

macOS

## ✔️ Possible solutions

First we should evaluate if a lock file should be part of the package or not.
Common sense is that it should not.
But I feel we are dealing with a different use case. We should study that use case, which is a distribution of a "binary" package, not a library.

Another problem is that our project use `yarn` and brew uses `npm`.
So we have a `yarn.lock`, in the root of the monorepo. Brew would only work with a `package-lock.json` as it uses `npm`.

tuler:
Bump dependencies.
After that I think we should release a 0.9.1, and try to understand what is going on with people's installation, while we don't have a better testing infrastructure, and fix #176


endersonmaia (approved):
tested it locally, and it's working, can't find out why CI is failing.

tuler:
I'll work on fixing the CI issue

tuler:
For some reason that I don't know yet the remote turbo cache entry for the `contracts` sub-project is invalid.
The cache key of the bogus entry is `fd1242d9137c4e1a`. I confirmed it's incorrect by linking my local build to the remote cache, and inspecting the cache file that turbo downloaded.

I added a commit to this PR adding a README file to the `contracts` project, changing its cache key.
Now the CI is passing.

I'll keep an eye on the issue, because there may some kind of build misconfiguration.

endersonmaia (approved):

Issue #245, opened 2023-10-08
Nobackend creates .sunodo dir with wrong ownership


gligneul:
## 🙂 Expected behavior

Create the dir with correct ownership, or don't create the dir.

## 🫠 Actual behavior

If I run `sunodo run --no-backend` without a `.sunodo` directory already created, this happens:

```
$ ll -a . | grep sunodo
drwxr-xr-x  3 root     root     4096 Oct  8 08:52 .sunodo
```

## 🧪 Minimal test case

In a clean sunodo project, run `sunodo run --no-backend`.

## 🌎 Environment

Tested with sunodo 0.9.0.
I'm using Debian with the native Docker engine.

PR #264, opened 2023-10-18
fix(cli): use env_file over environment at compose

### LINKED ISSUE #263: environments variables at .sunodo.env are overwritten by the ones defined at main compose environment: section
## 🙂 Expected behavior

When defining the `RD_EPOCH_DURATION` rollups-dispatcher configuration via `.sunodo.env`, I expect it to be defined inside the validator container.

## 🫠 Actual behavior

The `RD_EPOCH_DURATION` is not redefined accordingly, since it's defined at the `environment:` section of the validator service, and this overrides what's defined via `env_file:`

From: https://docs.docker.com/compose/compose-file/05-services/#env_file

> Environment variables declared in the [environment](https://docs.docker.com/compose/compose-file/05-services/#environment) section override these values. This holds true even if those values are empty or undefined.

## 🧪 Minimal test case

```
sunodo create --template=typescript my-ts-dapp
cd my-ts-dapp
sunodo build
echo "RD_EPOCH_DURATION=30" > .sunodo.env
echo "HELLO=world" >> .sunodo.env
sunodo run
# on another terminal
docker exec -ti <project-id>-validator-1 sh -c 'echo $HELLO'
world
docker exec -ti <project-id>-validator-1 sh -c 'echo $RD_EPOCH_DURATION'
86400
```

You'll see that `$RD_EPOCH_DURATION` remains the default value of `86400` defined at https://github.com/sunodo/sunodo/blob/%40sunodo/cli%400.9.3/apps/cli/src/node/docker-compose-dev.yaml#L117, although the `$HELLO` variable returned the valued defined.

## 🌎 Environment

```
 sunodo --version
@sunodo/cli/0.9.3 linux-x64 node-v20.8.0
```

## ✔️ Possible solutions

We could try using only `env_file`, and use the list order to guarantee that what's defined later is what's used.

From: https://docs.docker.com/compose/compose-file/05-services/#env_file

> env_file can also be a list. The files in the list are processed from the top down. For the same variable specified in two env files, the value from the last file in the list stands.

endersonmaia:
/close #263


apps/cli/src/node/docker-compose-dev.yaml
```diff
@@ -83,75 +83,8 @@ services:
...
-            ## uses redis
-            ## uses contract-address-file
-            ## uses chain-id
-
-            # graphql-server
-            ## uses postgres-config
-            GRAPHQL_HOST: 0.0.0.0
-            GRAPHQL_PORT: 4000
+        env_file:
+            - ${SUNODO_BIN_PATH}/node/default.env
```
oap75, line 87: We may get similar error on macos according to the issue #245

oap75 (changes requested):

apps/cli/src/node/docker-compose-dev.yaml
endersonmaia: Can you test this? I don't have a maOS.

oap75: Me either !
I think @tuler can help.

endersonmaia: I think this is not the same case from #245 , since it's not a volume mapping.

tuler: If I remove the docker volume mapping permission it does not work properly.

Strangely it does not work in non-verbose mode. But works in verbose mode.

endersonmaia: But this PR makes no changes to volumes. :thinking:

tuler:
Shouldn't we remove env variables in `run.ts` that have a constant value, and is already defined in `default.env`?

apps/cli/src/node/docker-compose-dev.yaml
tuler: Yeah, it may be something not introduced by this PR.
Topic for #267 and the removal of volume mapping in sunodo dir.

endersonmaia:
> Shouldn't we remove env variables in `run.ts` that have a constant value, and is already defined in `default.env`?

I'll take a look.

apps/cli/src/commands/run.ts
```diff
@@ -71,17 +71,13 @@ export default class Run extends Command {
             ANVIL_VERBOSITY: flags.verbose ? "--steps-tracing" : "--silent",
             BLOCK_TIME: blockInterval.toString(),
             BLOCK_TIMEOUT: (blockInterval + 3).toString(),
-            CHAIN_ID: "31337",
-            EPOCH_DURATION: epochDuration.toString(),
-            MNEMONIC: DEFAULT_DEVNET_MNEMONIC,
```
tuler, line 76: Remove unused DEFAULT_DEVNET_MNEMONIC variable from the code

tuler: Compilation broke

endersonmaia: removed too much :sweat: 

push forced on last fixup

tuler (approved):

oap75 (commented):
what if we use `--no-backend` option ? 
  
  the hostmode docker compose file ENVs will be replaced ?

endersonmaia:
> what if we use `--no-backend` option ?
> 
> the hostmode docker compose file ENVs will be replaced ?

`docker-compose-host.yaml` uses `environment: ` for only two variables, that should be used

To avoid it being replaced by `./sunodo.env` or `default.env`, I keep it like it is now.

We could use `env_file` everywhere, but that would require more attention about the ordering.

The way it is now, it's working and manageable, in case this become more complex, we should think on another approach to control this.

oap75 (approved):

PR #267, opened 2023-10-18
feat: change traefik config

### LINKED ISSUE #246: Mounting volume from sunodo bin directory is rejected by Docker
## 🙂 Expected behavior

During `sunodo run` we now mount a docker compose volume in the path of the sunodo binary installation.
Sunodo run should always work.

## 🫠 Actual behavior

Docker requires that the directory is the white list of the directories that can be mounted.
Which is my case was not, raising the following error:

```
Error response from daemon: Mounts denied: 
The path /opt/homebrew/Cellar/sunodo/0.9.0/libexec/lib/node_modules/@sunodo/cli/dist/node/traefik is not shared from the host and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> Resources -> File Sharing.
See https://docs.docker.com/desktop/mac for more info.
```

## 🧪 Minimal test case

Run `sunodo run` in a machine that doesn't have the sunodo bin dir in the while list.

## 🌎 Environment

macOS
Sunodo 0.9
Docker Desktop

## ✔️ Possible solutions

Setting up Docker to add the sunodo bin directory work.
But we should study a way to not need to map that volume

oap75:
/close #246 

This PR uses `docker provider` (traefik auto discovery) along with `compose labels` to define the traefik rules.

Finally, we can remove the volume and fix the issue.


endersonmaia:
It fails on my machine with:

```
dependency failed to start: container 54aef9b3-proxy-1 is unhealthy
```

oap75:
> It fails on my machine with:
> 
> ```
> dependency failed to start: container 54aef9b3-proxy-1 is unhealthy
> ```

Fixed ! 
Let's test again !

.changeset/pretty-turtles-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sunodo/cli": minor
```
tuler, line 2: As it does not change the behavior for the end user, and it's mostly a patch to a potential issue, maybe this should be a patch?

apps/cli/src/node/docker-compose-dev.yaml
```diff
@@ -159,14 +174,26 @@ services:
...
+                "--providers.docker=true",
+                "--providers.docker.endpoint=unix:///var/run/docker.sock",
+            ]
         ports:
-            - 8080:8080
+            - 8080:8088
+            - 8088:8080
         volumes:
-            - ${SUNODO_BIN_PATH}/node/traefik:/etc/traefik
+            - /var/run/docker.sock:/var/run/docker.sock
```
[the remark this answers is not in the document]
tuler: Yeah, this is not a good option. We may have problems with different docker installations, and privileged vs unprivileged modes.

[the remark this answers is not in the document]
endersonmaia: I think we should go the CLI arguments path, and avoid exposing the `docker.sock`, since this would be listening to any other container events running on the developer machine.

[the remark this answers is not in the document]
oap75: > I think we should go the CLI arguments path, and avoid exposing the `docker.sock`, since this would be listening to any other container events running on the developer machine.

We can use `[onstraints](https://doc.traefik.io/traefik/providers/docker/#constraints) to avoid this. 

But, the main question is: Are there any solutions that don't need to mount the `docker.sock` volume  and config files?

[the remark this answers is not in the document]
endersonmaia: Yes, CLI arguments or ENV variables.

https://doc.traefik.io/traefik/reference/static-configuration/cli/

https://doc.traefik.io/traefik/reference/static-configuration/env/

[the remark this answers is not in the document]
oap75: @endersonmaia There is no option to define all routes and services throw args nor ENVs. 

We should use one the traefik providers at first (we can define it via ENV or CMD). then we should configure the routes in the provider section. So, based on our scenario we have two options : 

1 - docker provider (we just need to mount docker.sock)
2 - file provider (we need to mount the config files --> yaml or toml format)


So, I think we can use docker format to remove the files, but one problem is the traefik should connect to docker.sock to read the labels + discover the IP of the containers.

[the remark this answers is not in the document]
endersonmaia: So the premises are:

-  routers, services and middleware cant be hold in the static configuration, but only via providers
- file provider is the solution to use, but we don't want to mount a local volume

While talking to Omid about this, we thought of creating a new service just to provision this configuration on a docker volume for the `proxy` service to use.

The static part could be on cli arguments, like this:

```
    command
          - --entrypoints.web.address=:8080
          - --entrypoints.traefik.address=:8088
          - --ping
          - --providers.file.directory=/etc/traefik/conf.d
          - --metrics.prometheus.addServicesLabels=true
```

And we manage to feed this `/etc/traefik//conf.d` directory.

This way we avoid mounting local host's volume and also avoid maintaining another container image.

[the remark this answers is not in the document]
oap75: image


image


image


image

apps/cli/src/node/docker-compose-dev.yaml
```diff
@@ -158,15 +158,28 @@ services:
         depends_on:
             validator:
                 condition: service_healthy
+            traefik-config-generator:
+                condition: service_completed_successfully
         healthcheck:
-            test: ["CMD", "traefik", "healthcheck"]
+            test: ["CMD", "traefik", "healthcheck", "--ping"]
```
endersonmaia, line 164: is this necessary? it was already working, right?

```suggestion
            test: ["CMD", "traefik", "healthcheck"]
```

endersonmaia (approved):

apps/cli/src/node/docker-compose-dev.yaml
oap75: I already tested it when I had enabled docker and it was not working. 

I'll test it again with new configuration.

oap75: @endersonmaia We should define it otherwise we will get this error : 


```
dependency failed to start: container sunodo-node-proxy-1 is unhealthy
```

image



with `--ping` all works:

apps/cli/src/node/docker-compose-traefik-config.yaml
```diff
@@ -0,0 +1,66 @@
...
+                            rule: "PathPrefix(`/explorer`)"
+                            service: explorer
+                        graphql_server:
+                            rule: "PathPrefix(`/graphql`)"
+                            service: graphql_server
+                        host-runner:
+                            rule: "PathPrefix(`/host-runner`)"
+                            middlewares:
+                                - "remove-host-runner-prefix"
+                            service: host-runner
```
endersonmaia, line 25: This should be only used when `--no-beackend` is used.

apps/cli/src/node/docker-compose-traefik-config.yaml
```diff
@@ -0,0 +1,66 @@
...
+                                servers:
+                                    - url: "http://explorer:3000"
+                        graphql_server:
+                            loadBalancer:
+                                servers:
+                                    - url: "http://validator:4000/graphql"
+                        host-runner:
+                            loadBalancer:
+                                servers:
+                                    - url: "http://validator:5004"
```
endersonmaia, line 56: This should be only used when `--no-beackend` is used.

endersonmaia (changes requested):

apps/cli/src/node/docker-compose-traefik-config.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+                    middlewares:
+                        remove-explorer-api-prefix:
+                            replacePathRegex:
+                                regex: "^/explorer-api/(.*)"
+                                replacement: "/$1"
+                        remove-host-runner-prefix:
+                            replacePathRegex:
+                                regex: "^/host-runner/(.*)"
+                                replacement: "/$1"
```
endersonmaia, line 38: This should be only used when `--no-beackend` is used.

endersonmaia (changes requested):

endersonmaia (commented):
You solved that host-runner is not provided when not needed but introduced code duplication.

May I send a patch with a different approach?

oap75:
> You solved that host-runner is not provided when not needed but introduced code duplication.
> 
> May I send a patch with a different approach?

Actually I have another fixup. pls let me send it to review first.

endersonmaia (dismissed):
Made some comments about the port redirections, but it's not a blocker.

apps/cli/src/node/docker-compose-dev.yaml
```diff
@@ -158,15 +158,28 @@ services:
...
             interval: 5s
             timeout: 1s
             retries: 5
+        command:
+            [
+                "--ping=true",
+                "--entryPoints.web.address=:8088",
+                "--entryPoints.traefik.address=:8080",
```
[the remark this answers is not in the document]
oap75: During the `PING health check`, we need to assign the traefik address (management port) to `8080` because there is no option to define any ports specific (actually there is and I have tested it it did not work) for the `healthcheck ping command` and the default port for it is `8080` for it. 

That's why I have changed the management port and the web-address port locally. 

But, changed the docker compose mapping port to work as same as before. (localhost:8080 ---> web-addresses)

oap75:
@tuler I think the PR is ready to merge. Any thoughts?

tuler:
I’ll review it beginning of next week

apps/cli/src/node/docker-compose-dev.yaml
```diff
@@ -91,30 +91,78 @@ services:
...
+                "--metrics.prometheus=true",
+                "--metrics.prometheus.addServicesLabels=true",
+                "--providers.file.directory=/etc/traefik/conf.d",
+                "--providers.file.watch=true"
+            ]
         ports:
-            - 8080:8080
+            - 8080:8088
+            - 8088:8080
```
tuler, line 113: Is this reverted? Can we not expose the management port?

apps/cli/src/node/docker-compose-host.yaml
```diff
@@ -22,3 +22,28 @@ services:
             ENABLE_SUNODO_NODE: true
             # advance-runner
             SNAPSHOT_ENABLED: false
+
+    traefik-config-generator:
```
tuler, line 26: I see we are using the same service-name for all sub-composes to augment the environment, is that correct?

endersonmaia: yes, the idea was to keep the configuration close to the service it configures

apps/cli/src/node/docker-compose-dev.yaml
endersonmaia: I'll remove the `:8088` port.

endersonmaia (approved):

tuler (approved):

PR #268, opened 2023-10-18
fix: anvil nightly version

oap75:


tuler:
Nice, please rebase and add a changeset

endersonmaia (changes requested):
I miss a changeset, I think it should be a patch version.

endersonmaia (approved):

tuler:
I meant a changeset for the anvil subproject.
We need to first release the anvil package/docker image.
Updating of the CLI comes later

PR #270, opened 2023-10-18
chore(anvil): fix anvil changeset

tuler:
I know this also fixes the build of the entire project, but the main thing of the changeset is to introduce a new release of the anvil image. The changeset of the CLI comes later.


tuler:
I'm using minor just because.
The anvil versioning itself is already messed up.

endersonmaia (approved):

PR #271, opened 2023-10-18
feat(rollups-node): supress messages from validator services

endersonmaia:


packages/rollups-node/s6-overlay/stage2-hook.sh
```diff
@@ -1,8 +1,8 @@
 #!/command/with-contenv sh
+set -e
 
 # check whether to use host-runner or not
 if [ ! -z ${ENABLE_SUNODO_NODE+x} ]; then
-    /command/s6-echo "host-runner: ENABLE_SUNODO_NODE is set, running in host mode..."
```
tuler, line 5: There is a related issue which is showing the host-runner is running and at what address.
I did not include it in the PR #231

tuler (approved):

packages/rollups-node/s6-overlay/stage2-hook.sh
endersonmaia: Should I fix this in this PR?

PR #276, opened 2023-10-20
feat: change prompt config and remove docker volume from it

### LINKED ISSUE #246: Mounting volume from sunodo bin directory is rejected by Docker

oap75:
/close #246 
This PR will remove the 'prompt' Docker volume and use environment variables instead.

### DIFF
--- a/apps/cli/src/node/docker-compose-dev.yaml
+++ b/apps/cli/src/node/docker-compose-dev.yaml
@@ -183,15 +183,24 @@ services:
 
     prompt:
         image: debian:bookworm-slim
+        environment:
+            PROMPT_TXT_01_ANVIL: "Anvil running at http://localhost:8545"
+            PROMPT_TXT_02_GRAPHQL: "GraphQL running at http://localhost:8080/graphql"
+            PROMPT_TXT_03_INSPECT: "Inspect running at http://localhost:8080/inspect/"
+            PROMPT_TXT_04_EXPLORER: "Explorer running at http://localhost:8080/explorer/"
+            PROMPT_TXT_10_EXIT: "Press Ctrl+C to stop the node"
         command:
-            [
-                "sh",
-                "-c",
-                "ls | sort | xargs cat; trap exit INT TERM; sleep infinity & wait",
-            ]
+            - /bin/bash
+            - -c
+            - |
+                for txt_file in $${!PROMPT_TXT_*}
+                do
+                    declare -n txt_content=$$txt_file
+                    _txt_file=$${txt_file:11}
+                    echo "$$txt_content"
+                done
+                trap exit INT TERM; sleep infinity & wait
         working_dir: /prompt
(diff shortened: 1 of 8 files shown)


apps/cli/src/node/docker-compose-dev.yaml
```diff
@@ -183,15 +183,24 @@ services:
...
+            - /bin/bash
+            - -c
+            - |
+                for txt_file in $${!PROMPT_TXT_*}
+                do
+                    declare -n txt_content=$$txt_file
+                    _txt_file=$${txt_file:11}
+                    echo "$$txt_content" > /prompt/$${_txt_file,,}.txt
+                done
+                ls | sort | xargs cat; trap exit INT TERM; sleep infinity & wait
```
endersonmaia, line 202: Since we don't need to transform this into files, we could just print the messages.

```suggestion
                    echo "$$txt_file"
                done
```

Not sure if we need to `wait`.

If that's the case:

```suggestion
                    echo "$$txt_file"
                done
                trap exit INT TERM; sleep infinity & wait
```

endersonmaia (changes requested):
I think you could use this PR to add the `/host-runner` endpoint to the messages when `--no-backend` is used.

apps/cli/src/node/docker-compose-host.yaml
```diff
@@ -43,3 +43,7 @@ services:
                             loadBalancer:
                                 servers:
                                     - url: "http://validator:5004"
+
+    prompt:
+        environment:
+            PROMPT_TXT_05_HOST_RUNNER: "Host_Runner running at http://localhost:5004/host-runner"
```
endersonmaia, line 49:
```suggestion
            PROMPT_TXT_05_HOST_RUNNER: "Host Runner running at http://localhost:5004/host-runner"
```

endersonmaia (changes requested):

PR #281, opened 2023-10-24
fix(ci): container release with latest tag

### LINKED ISSUE #280: released container artifacts has wrong latest tags
## 🙂 Expected behavior

sunodo container releases should point the `:latest` tag to that latest release.

## 🫠 Actual behavior

- `sunodo/rollups-node:latest` - Oct 3, 2023 at 5:11 pm
- `sunodo/rollups-node:0.5.1` - Oct 24, 2023 at 3:25 pm

## ✔️ Possible solutions

- [x] Fix the CI to also release the latest tag for all container artifacts
- [ ] Point the latest tag to all already released container artifacts and not wait for the next CI run

endersonmaia:
/close #280

### DIFF
--- a/.github/workflows/anvil.yaml
+++ b/.github/workflows/anvil.yaml
@@ -39,6 +39,7 @@ jobs:
                       ghcr.io/sunodo/anvil
                   tags: |
                       type=raw,value=${{ steps.package-version.outputs.PACKAGE_VERSION }},enable=${{ github.event_name == 'push' }}
+                      type=raw,value=latest,enabled=${{ github.event_name == 'push' }}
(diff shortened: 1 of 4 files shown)


tuler:
This strategy is not the correct one, because it can set version `1.2.3` as the latest while there is already a `1.3.0`.

endersonmaia:
> This strategy is not the correct one, because it can set version `1.2.3` as the latest while there is already a `1.3.0`.

We don't have any maintenance release now, our releases are always increasing the version.

What do you think applying this patch, and add an issue to track this scenário you described?

PR #296, opened 2023-11-14
Start command (patch namespace)

tuler:
This shows that this strategy of patching templates generated by helm is not very nice.
Traefik when installed in a different namespace, needs more patches, especially related to RBAC.

### DIFF
--- a/apps/cli/src/k8s/traefik/templates/dashboard-ingressroute.yaml
+++ b/apps/cli/src/k8s/traefik/templates/dashboard-ingressroute.yaml
@@ -4,11 +4,11 @@ apiVersion: traefik.io/v1alpha1
 kind: IngressRoute
 metadata:
   name: traefik-dashboard
-  namespace: default
+  namespace: sunodo-31337
   annotations:
   labels:
     app.kubernetes.io/name: traefik
-    app.kubernetes.io/instance: traefik-default
+    app.kubernetes.io/instance: traefik-sunodo-31337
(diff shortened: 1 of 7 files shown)


endersonmaia:
> This shows that this strategy of patching templates generated by helm is not very nice.

For the namespace, can't you define during the `helm template` with `--namespace` helm's flag?

Only what needs to be defined at runtime should be "patched" using k8s client patch support.

tuler:
> For the namespace, can't you define during the `helm template` with `--namespace` helm's flag?

That’s what I did for this templates, to exemplify.

> Only what needs to be defined at runtime should be "patched" using k8s client patch support.

All the changes in the example PR should be patched in runtime. I opened the PR just to show it’s not trivial.

endersonmaia:
> > For the namespace, can't you define during the `helm template` with `--namespace` helm's flag?
> 
> That’s what I did for this templates, to exemplify.
> 
> > Only what needs to be defined at runtime should be "patched" using k8s client patch support.
> 
> All the changes in the example PR should be patched in runtime. I opened the PR just to show it’s not trivial.

I don't get it. 

You shouldn't be patching that YAML manifest files itself, but loading it and using some mechanism to patch it (in-memory) before applying it into k8s API. 

I'm not sure how to do it from javascript, I'm assuming it's possible.

tuler:
This PR is just to show the in-memory changes I’d need to do in the traefik templates.

PR #298, opened 2023-11-27
Get rid of machine emulator image in favor of custom server manager image

alexmikhalevich:
This PR introduces two changes highlighted in the sunodo channel.

**Usage of custom emulator images**

When there is a need to use a custom emulator image, it is unfortunate that we can use config files only for tag specification. We propose using build args for the registry & org as well.

**Using server manager image instead of machine emulator**

Stored machines are not compatible with all the emulator versions, only with the one it was generated by. Thus, for improved correctness, the emulator image should not be used directly. We propose using the server manager image instead. This way it would be easy to ensure that the versions used are in a consistent state.

Please, let me know what you think about these changes.


endersonmaia:
The changes seem harmless, but I don't get how can I use this changes from the `sunodo build` command.

Could you exemplify what's the intended use case?

This PR may be a preparation for another PR with a complete feature, if that's the case, we can merge.

tuler:
I understand that this PR facilitates the job for the sunodo developer to change the emulator. Not necessarily an end user to do that, because this is an advanced use case (to change the emulator version).

endersonmaia (approved):

PR #301, opened 2023-12-01
feat(contracts): add PayableInputBox

felipeargento:
Hello Sunodo, I think this is my first PR here :)

Dropping in the first draft of a PayableInputBox implementation with its factory. This is a rough cut to kickstart our discussions, so we can agree on the interface/implementation.

Here’s the summary:

**PayableInputBox**: It accepts an ERC20 tokens as payment for inputs. You add an input, it deducts tokens from your account and sends it to the "authority address" (which should probably not be name like this) and then forwards the input to a base InputBox contract.

Factor: There's also a factory contract to churn out these PayableInputBoxes with custom settings.

Key things to debate:

**Customizable Parameters**:

Owners can tweak stuff like the token price per input, who gets the tokens (authority address), and which InputBox we’re forwarding to. Owners having this much control is quite weird, they can out of nowhere bump the price of adding inputs and hold everyone as a hostage. We could consider adding a "delay" here, so that users can protect themselves from malicious dapp owners. Or remove ownership all together - and make the dapp deal with that explicitly. What do you think?

**addInput receives dapp address**: 
I'm thinking that maybe the dapp address should go to the factory, since each dapp will likely have different pricing for inputs. But on this naive implementation (where all inputs cost the same), maybe they'll want to share a contract?

Anyway, the implementation made the design seems a bit off. I'll reconsider it. Let me know if you have any suggestions.

ps: I haven't contribute to the code in a long time, so let me know if the file is in the correct place / the commit is following the standard.

Also, is configuration working as intended? I can't run `forge` or `yarn prettier` inside the folder.


tuler:
Regarding the price mutability I think the way is to allow an authority to change and becoming effect after some time.

PR #305, opened 2023-12-08
Multi-dapp node

tuler:
Implementation of the management API proposed at https://github.com/cartesi/rollups-node/pull/184

This version does not persist managed applications yet.
The execution of nodes is a simple spawn of a child process.
Stdio of the child inherits the stdio of the parent process.


packages/rollups-node/s6-overlay/s6-rc.d/sunodo-node/up
endersonmaia, line 1: this filename should be `run`, since it's of type `longrun`

just replicate the file structure of any other onlgrun service

packages/rollups-node/s6-overlay/s6-rc.d/sunodo-node/up
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+sunodo-node --port 3000
```
endersonmaia, line 2: To use unprivileged `cartesi` user.

The timeout configuration could be different depending on how much time it takes to start this node version.

```suggestion
#!/command/with-contenv sh
exec s6-notifyoncheck -w 500 -t 500 -n 5 \
    s6-setuidgid cartesi sunodo-node --port 3000
```

tuler:
Closing this as the new rollups-node is multi-application.
Even if we still miss an API for application management.

PR #328, opened 2023-12-19
feat: call homebrew-tap Dispatch workflow

omidasadpour:
This PR added a `Repository Dispatcher` section to call the `sunodo/homebrew-tap` workflows to automate the NPM package update process. 

We just need to generate a `Fine-grained personal access tokens` and set it on github secrets.


tuler:
Let's discuss another way with the new Cartesi CLI

Issue #300, opened 2023-12-01
Prepare for next cartesi emulator SDK release


endersonmaia:
## 📚 Context

Prepare for next emultator SDK release.

## 📈 Subtasks

- [x] ROM got integrated into the emulator (There is no more: `--rom-image`)
- [x] use xgenext2fs binary when version is 1.5.3 instead of genext2fs
- [x] `--rollup` option got removed in favor of being the default. Use `--no-rollup` to disable it instead. (examlpe: https://github.com/sunodo/sunodo/blob/320df7b00c7d2ab80a2beeea782f168188755101/apps/cli/src/commands/shell.ts#L44)
- [x] `--append-rom-bootargs` got renamed to `--append-bootargs` (example: https://github.com/sunodo/sunodo/blob/320df7b00c7d2ab80a2beeea782f168188755101/apps/cli/src/commands/shell.ts#L49)
- [x] `tools.tar.gz` is being removed use `.deb` instead (example: https://github.com/sunodo/sunodo-templates/blob/main/go/Dockerfile#L40)

edubart:
- [ ] New append init commands commands (executed as root privilege before the entrypoint commands). Ideally Sunodo should provide a way to customize the init, because some dapps may want to change folder/devices permissions as root before running the entrypoint.
>   --append-init=<string>
    append a command to machine's init script to be executed with root privilege.
    The command is executed on boot after mounting flash drives and before running the entrypoint.
    You can pass this option multiple times.

>  --append-init-file=<filename>
    like --append-init, but use contents from a file.

- [ ] New append entrypoint commands (executed as dapp privilege before the final dapp command). Ideally Sunodo should provide a way to add customize entrypoint, because some dapps may want to execute additional commands as dapp user before running the final entrypoint command.

>   --append-entrypoint=<string>
    append a command to machine's entrypoint script to be executed with dapp privilege.
    The command is executed after the machine is initialized and before the final entrypoint command.
    You can pass this option multiple times.

>  --append-entrypoint-file=<filename>
    like --append-entrypoint, but use contents from a file.

- [ ] `--append-bootargs` was renamed, ideally Sunodo should provide a way to add extra bootargs, because some dapps may want to enable/disable kernel features before booting.
- [ ] You can now customize the entrypoint user with `--append-init=USER=root`, this replaces the `--append-rom-bootargs='single=yes'`.
- [ ] You can now customize the entrypoint working directory with `--append-init=WORKDIR=/tmp`.
[...]

Issue #193, opened 2023-09-06
Custom kernel, bootargs and flash drives


edubart:
## 📚 Context

I am experimenting a dapp where I need a custom kernel, custom bootargs and extra flash drives. Currently there is no way to do this with Sunodo.

## ✔️ Solution

Would be nice if Sunodo could allow me to specify a different kernel image, different kernel headers, extra bootargs and extra flash drive configurations.

A plus would be a label to pass custom arguments to the `cartesi-machine` command, this would allow to workaround some of these limitations without Sunodo having official support yet.

tuler:
Regarding the kernel, it's part of the `sunodo/sdk:0.2.0` image, which is built at https://github.com/sunodo/sunodo/tree/main/packages/sdk

Sunodo `build` looks for that image, which can be customized using the Docker label `io.sunodo.sdk_version`.
If you define in the Dockerfile:

```
LABEL io.sunodo.sdk_version=my-custom-kernel
```

It will use the docker image `sunodo/sdk:my-custom-kernel`.

This could still be improved to replace the entire docker image name, not only the tag.

tuler:
I'm considering to move away from the configuration through Docker labels and introducing a sunodo.yaml configuration file.

Any thoughts?

edubart:
> Regarding the kernel, it's part of the sunodo/sdk:0.2.0

I am doing that way now, it works. But looks like this `LABEL` config does not reflect `sunodo shell` command.

> I'm considering to move away from the configuration through Docker labels and introducing a sunodo.yaml configuration file.

I think it makes sense, this would open the possibility to customize other commands too, not just `sunodo build`. Like it could have also configs for `sunodo run`, and others

tuler:
> But looks like this `LABEL` config does not reflect `sunodo shell` command.

You're right, it's a known issue. See comment https://github.com/sunodo/sunodo/blob/26169e278a3004d6139b82f4aa05560f457ace76/apps/cli/src/commands/shell.ts#L34

gbarros:
I think it makes a lot of sense to have a `sunodo.yaml`; many CLIs have it, and I feel it's a natural progression. 
I would add that support to define env vars is nice to have =]

gligneul:
[...]

PR #333, opened 2023-12-22
Cartesi Machine SDK 0.17.1

endersonmaia:
Related to #300


tuler:
`sunodo build` started to work.

Using templates from https://github.com/sunodo/sunodo-templates/pull/37

packages/sdk/Dockerfile
```diff
@@ -34,26 +41,28 @@ FROM builder as su-exec
 
 # v0.2 -> f85e5bde1afef399021fbc2a99c837cf851ceafa
 WORKDIR /usr/local/src
-RUN <<EOF
-set -e
-curl -sL https://github.com/ncopa/su-exec/archive/f85e5bde1afef399021fbc2a99c837cf851ceafa.tar.gz | tar --strip-components=1 -zxvf -
-make
-EOF
+ADD --keep-git-dir=true https://github.com/ncopa/su-exec.git#f85e5bde1afef399021fbc2a99c837cf851ceafa /usr/local/src
```
tuler, line 44: any reason for the `--keep-git-dir`?

packages/sdk/Dockerfile
```diff
@@ -67,18 +76,21 @@ ENV LANGUAGE en_US:en
 
 COPY entrypoint.sh /usr/local/bin/
 COPY retar /usr/local/bin/
-COPY --from=genext2fs /usr/local/bin/genext2fs /usr/bin/
+COPY --from=genext2fs /usr/local/bin/xgenext2fs /usr/bin/
```
tuler, line 79: why is this not going to `/usr/local/bin`?

tuler (commented):
We should separate a PR for SDK 0.4. Which is very close to approval.
And publish it.

tuler:
Separated CLI changes to PR https://github.com/sunodo/sunodo/pull/431

tuler:
Should we add the rollups node in the SDK, and use the SDK in the CLI compose?
This can facilitate #193

packages/sdk/Dockerfile
endersonmaia: In this su-exec case, no particular reason.

packages/sdk/Dockerfile
endersonmaia: It will work anyway, but for better semantics and consistency, we could put at `/usr/local/bin`.

endersonmaia:
> Should we add the rollups node in the SDK, and use the SDK in the CLI compose? This can facilitate #193

If we follow this path, should `sunodo/sdk` derive `FROM cartesi/rollups-node` instead?

tuler:
> If we follow this path, should `sunodo/sdk` derive `FROM cartesi/rollups-node` instead?

If we follow this path, yes.

What are the pros and cons?
Supposing we have a flag in `sunodo run`, which can be hidden from regular users, that allows a custom validator image.

If SDK includes node:
- In case of a new emulator version we need to update both the SDK and CLI (to change the default version)
- In case of a new node we need to update both the SDK and CLI as well
- In case of a advanced user he needs to create a custom SDK image and use that in `sunodo run`

if SDK does not include node:
- In case of a new emulator version we need to update both the SDK and CLI (same thing)
- In case of a new node we need to update only the CLI (easier)
- In case of a advanced user he needs to create a custom SDK image and use that in `sunodo run` (same thing)

Right?

So no advantages I guess. Just need to allow sunodo run to use a custom image.

endersonmaia:
> So no advantages I guess. Just need to allow sunodo run to use a custom image.

Is there already an issue to track that?

tuler:
> Is there already an issue to track that?

Not specifically that, but we have #193

packages/sdk/Dockerfile
tuler: Can I remove all `--keep-git-dir` from this Dockerfile?
This and the one from genext2fs

endersonmaia: What your concern here?

I think you could remove it. 

Some build processes use the git workspace to gather information about versioning and stuff, don't know if that's the case here.

tuler: I don't think that's the case for theses builds.
Just wanted to remove an explicit option if we don't explicitly need it.

packages/sdk/Dockerfile
```diff
@@ -3,27 +3,34 @@ ARG BASE_IMAGE
...
+    ca-certificates \
+    curl \
+    libarchive-dev \
+    libtool
 rm -rf /var/lib/apt/lists/*
 EOF
 
 FROM builder as genext2fs
 
+# main -> a7abe9ecaeb376babc4148c1122dfb9103539d2c
```
endersonmaia, line 29: Let's use the tagged commit

`v1.5.3 -> 1c1992153eb77cc59a8447c631b141c955d37c2a `

tuler: The tagged commit does not have the latest fix @mpolitzer did

endersonmaia: maybe we should ask him to make a release

endersonmaia: but ok, if it doesn't have any other change, the tag will eventually point to the same commit

endersonmaia: anyway, the branch name on @mpolitzer's fork is `cartesi`.

tuler: I don't know if he has more commits to do to make it "1.5.4". Isn't there a `--version` that needs a change?

endersonmaia: probably https://github.com/cartesi/genext2fs/blob/cartesi/configure.ac#L5

tuler: https://github.com/cartesi/genext2fs/commit/1c1992153eb77cc59a8447c631b141c955d37c2a

tuler (approved):

PR #335, opened 2024-01-03
Fix token initial mint

tuler:
I was minting the initial amount to the `msg.sender` of the deployer of the contract.
This was causing a problem with the deterministic deployment, because the `msg.sender` in that case was the smart contract that actually does the deployment, which was the account used by https://github.com/safe-global/safe-singleton-factory.
This changes the recipient of the initial mint to the initial authority of the contract, which is the one that has minting power anyway, so it works just fine.

### DIFF
--- a/.changeset/fifty-singers-nail.md
+++ b/.changeset/fifty-singers-nail.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/token": minor
+---
+
+change initial mint to the initial authority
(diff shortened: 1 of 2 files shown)


endersonmaia (changes requested):
You need to rebase this, considering just the last commit, right?

endersonmaia (approved):

PR #353, opened 2024-01-26
Improve reproducible build

### LINKED ISSUE #330: sunodo build not reproducible
## 🙂 Expected behavior

Two builds from the same template should have the same hash.

## 🫠 Actual behavior

When building two applications from the same template, I get different hashes.

## 🧪 Minimal test case

```shell
#!/usr/bin/env bash

_back=$(pwd)

sunodo --version

cd $(mktemp -d)
sunodo create --template=typescript test-a
cd test-a
sunodo build
cd $_back

cd $(mktemp -d)
sunodo create --template=typescript test-b
cd test-b
sunodo build
cd $_back
```

Output:

```
❯ ./test.sh
@sunodo/cli/0.10.3 linux-x64 node-v20.10.0
✔ Application created at /tmp/tmp.ItdqDS3VYo/test-a
...
 => => writing image sha256:e577d7b8bb804437d3dff120b8076c94cd2c3c2d17d8ce836544cf93cd99343b                     0.0s
...
Manual yield rx-accepted (0x100000000 data)
Cycles: 920671735
920671735: 560e236bde532fb4018e0292f066f330d7750179e0336346ba84c9d0f75bb9b1
Storing machine: please wait
✔ Application created at /tmp/tmp.YzfRaYJ1mc/test-b
...
 => => writing image sha256:e577d7b8bb804437d3dff120b8076c94cd2c3c2d17d8ce836544cf93cd99343b                     0.0s
...
Manual yield rx-accepted (0x100000000 data)
Cycles: 920671735
920671735: 439141771b8b7f7c483514b2df72c9ade28a5cc9464cdb53c515286734757e2b
Storing machine: please wait

```

Using diffoscope I get information about the difference between them:

```shell
❯ diffoscope /tmp/tmp.ItdqDS3VYo/test-a/.sunodo/image.ext2 /tmp/tmp.YzfRaYJ1mc/test-b/.sunodo/image.ext2
--- /tmp/tmp.ItdqDS3VYo/test-a/.sunodo/image.ext2
+++ /tmp/tmp.YzfRaYJ1mc/test-b/.sunodo/image.ext2
├── image.ext2.tar
│ ├── file list
│ │ @@ -1,16 +1,16 @@
│ │  drwxr-xr-x   0        0        0        0 1970-01-01 00:00:00.000000 ./
│ │ --rwxr-xr-x   0        0        0        0 2023-12-21 10:14:09.000000 ./.dockerenv
│ │ +-rwxr-xr-x   0        0        0        0 2023-12-21 10:14:26.000000 ./.dockerenv
...
│ │ -drwxr-xr-x   0        0        0        0 2023-12-21 10:14:09.000000 ./dev/
│ │ --rwxr-xr-x   0        0        0        0 2023-12-21 10:14:09.000000 ./dev/console
│ │ -drwxr-xr-x   0        0        0        0 2023-12-21 10:14:09.000000 ./dev/pts/
│ │ -drwxr-xr-x   0        0        0        0 2023-12-21 10:14:09.000000 ./dev/shm/
│ │ -drwxr-xr-x   0        0        0        0 2023-12-21 10:14:09.000000 ./etc/
│ │ +drwxr-xr-x   0        0        0        0 2023-12-21 10:14:26.000000 ./dev/
│ │ +-rwxr-xr-x   0        0        0        0 2023-12-21 10:14:26.000000 ./dev/console
│ │ +drwxr-xr-x   0        0        0        0 2023-12-21 10:14:26.000000 ./dev/pts/
│ │ +drwxr-xr-x   0        0        0        0 2023-12-21 10:14:26.000000 ./dev/shm/
│ │ +drwxr-xr-x   0        0        0        0 2023-12-21 10:14:26.000000 ./etc/
...
│ │ --rwxr-xr-x   0        0        0        0 2023-12-21 10:14:09.000000 ./etc/hostname
│ │ --rwxr-xr-x   0        0        0        0 2023-12-21 10:14:09.000000 ./etc/hosts
│ │ +-rwxr-xr-x   0        0        0        0 2023-12-21 10:14:26.000000 ./etc/hostname
│ │ +-rwxr-xr-x   0        0        0        0 2023-12-21 10:14:26.000000 ./etc/hosts
...
│ │ -lrwxrwxrwx   0        0        0        0 2023-12-21 10:14:09.000000 ./etc/mtab -> /proc/mounts
│ │ +lrwxrwxrwx   0        0        0        0 2023-12-21 10:14:26.000000 ./etc/mtab -> /proc/mounts
...
│ │ --rwxr-xr-x   0        0        0        0 2023-12-21 10:14:09.000000 ./etc/resolv.conf
│ │ +-rwxr-xr-x   0        0        0        0 2023-12-21 10:14:26.000000 ./etc/resolv.conf
...
```

## 🌎 Environment

`@sunodo/cli/0.10.3 linux-x64 node-v20.10.0`

## ✔️ Possible solutions

Ensure that files' timestamps get normalized before ext2 generation.

endersonmaia:
This PR will add `undocker` to the sunodo/sdk image, so it solve the problem reported at #330, where the `docker container create` step introduces files that makes final ext2 rootfs not reproducible.

/close #330

### DIFF
--- a/.changeset/modern-mirrors-shake.md
+++ b/.changeset/modern-mirrors-shake.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/sdk": minor
+---
+
+add undocker to extract rootfs from container image
--- a/apps/cli/src/commands/build.ts
+++ b/apps/cli/src/commands/build.ts
@@ -3,6 +3,7 @@ import bytes from "bytes";
 import { execa } from "execa";
 import fs from "fs-extra";
 import path from "path";
+import semver from "semver";
 import tmp from "tmp";
 
 type ImageBuildOptions = {
@@ -111,6 +112,22 @@ export default class BuildApplication extends Command {
 
         if (!info.entrypoint && !info.cmd) {
             throw new Error("Undefined image ENTRYPOINT or CMD");
+        }
+
+        // fail if using unsupported sdk version
+        if (!semver.valid(info.sdkVersion)) {
+            this.warn("sunodo sdk version is not a valid semver");
+        } else if (semver.lt(info.sdkVersion, SUNODO_DEFAULT_SDK_VERSION)) {
+            throw new Error(`Unsupported sunodo sdk version.
+
+Make sure you defined \`io.sunodo.sdk_version\` label at your Dockerfile.
+
+It shoud be >= ${SUNODO_DEFAULT_SDK_VERSION}.
+
+Eg.:
+
+LABEL io.sunodo.sdk_version=${SUNODO_DEFAULT_SDK_VERSION}
+`);
         }
 
         // warn for using default values
--- a/apps/cli/src/commands/build.ts
+++ b/apps/cli/src/commands/build.ts
@@ -1,10 +1,9 @@
 import { Command, Flags } from "@oclif/core";
-import fs from "fs-extra";
-import os from "os";
-import path from "path";
-import { execa } from "execa";
-import tmp from "tmp";
 import bytes from "bytes";
+import { execa } from "execa";
+import fs from "fs-extra";
+import path from "path";
+import tmp from "tmp";
 
 type ImageBuildOptions = {
     target?: string;
@@ -27,11 +26,16 @@ const CARTESI_DEFAULT_RAM_SIZE = "128Mi";
 
 const SUNODO_LABEL_PREFIX = "io.sunodo";
 const SUNODO_LABEL_SDK_VERSION = `${SUNODO_LABEL_PREFIX}.sdk_version`;
-const SUNODO_DEFAULT_SDK_VERSION = "0.2.0";
+const SUNODO_DEFAULT_SDK_VERSION = "0.3.0";
 
-const SUNODO_DEFAULT_MACHINE_SNAPSHOT_PATH = path.join(".sunodo", `image`);
-const SUNODO_DEFAULT_TAR_PATH = path.join(".sunodo", `image.tar`);;
-const SUNODO_DEFAULT_EXT2_PATH = path.join(".sunodo", `image.ext2`);;
+const SUNODO_PATH = path.join(".sunodo");
+const SUNODO_DEFAULT_MACHINE_SNAPSHOT_PATH = path.join(SUNODO_PATH, "image");
+const SUNODO_DEFAULT_DOCKER_TAR_PATH = path.join(
+    SUNODO_PATH,
+    "docker-image.tar",
+);
+const SUNODO_DEFAULT_TAR_PATH = path.join(SUNODO_PATH, "image.tar");
+const SUNODO_DEFAULT_EXT2_PATH = path.join(SUNODO_PATH, "image.ext2");
 
 export default class BuildApplication extends Command {
     static summary = "Build application.";
@@ -63,7 +67,7 @@ export default class BuildApplication extends Command {
      * Build DApp image (linux/riscv64). Returns image id.
      * @param directory path of context containing Dockerfile
      */
-    private async buildDAppImage(options: ImageBuildOptions): Promise<string> {
+    private async buildImage(options: ImageBuildOptions): Promise<string> {
         const buildResult = tmp.tmpNameSync();
         this.debug(
             `building docker image and writing result to ${buildResult}`,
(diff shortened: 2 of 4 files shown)


packages/sdk/Dockerfile
```diff
@@ -61,6 +71,8 @@ COPY entrypoint.sh /usr/local/bin/
 COPY retar /usr/local/bin/
 COPY --from=genext2fs /usr/local/bin/genext2fs /usr/bin/
 COPY --from=su-exec /usr/local/src/su-exec /usr/local/bin/
+COPY --from=undocker /undocker/undocker /usr/local/bin/
+RUN mkdir -p /tmp/.sunodo && chmod 1777 /tmp/.sunodo
```
tuler, line 75: Do we need to create this directory here?

.changeset/modern-mirrors-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sunodo/sdk": patch
```
tuler, line 2: Let's bump the minor instead?

apps/cli/src/commands/build.ts
```diff
@@ -27,11 +26,12 @@ const CARTESI_DEFAULT_RAM_SIZE = "128Mi";
 
 const SUNODO_LABEL_PREFIX = "io.sunodo";
 const SUNODO_LABEL_SDK_VERSION = `${SUNODO_LABEL_PREFIX}.sdk_version`;
-const SUNODO_DEFAULT_SDK_VERSION = "0.2.0";
+const SUNODO_DEFAULT_SDK_VERSION = "0.2.1";
```
tuler, line 29: Related to bumping the minor instead

apps/cli/src/commands/build.ts
```diff
@@ -132,36 +132,74 @@ export default class BuildApplication extends Command {
...
     ): Promise<void> {
-        // create container
-        const { stdout: cid } = await execa("docker", [
-            "container",
-            "create",
-            "--platform",
-            "linux/riscv64",
+        // delete tar if exists
+        if (fs.existsSync(path))
+            fs.unlinkSync(path);
```
tuler, line 142: Do we need this? I already do a `fs.emptyDir(".sunodo")` in the beginning of the build process

endersonmaia: It was just to remove the `docker-image.tar` that I consider an intermediary.

But thinking about it, we could remove all except `./sunodo/image/` (used for `run`) and `.sunodo/image.ext2`  (used for `shell`).

tuler: Currently `.sunodo` is cleared in the beginning of the build process

endersonmaia: Yes, and the build result is kept until the next build, and these other files are kept there, just using space.

So I think we could remove not just the `./sunodo/docker-image.tar` but also the `./sunoso/image.tar`.

packages/sdk/Dockerfile
endersonmaia: Yes, the build failed without it.

apps/cli/src/commands/build.ts
tuler: Yeah, I intentionally left the image.tar around mainly for debug reasons.

endersonmaia: So I prefer removing the `.sunodo/docker-image.tar` since it wouldn't help with debugging from the POV of the generated `image.ext2`.

apps/cli/src/commands/build.ts
```diff
@@ -132,39 +132,67 @@ export default class BuildApplication extends Command {
...
-        tarPath: string,
-    ): Promise<string> {
+    private async getBuildContainer(image: string): Promise<string> {
         const { stdout: cid } = await execa("docker", [
             "container",
             "run",
             "--detach",
             "--init",
-            `--volume=${tarPath}:/tmp/${SUNODO_DEFAULT_TAR_PATH}`,
+            `--volume=./${SUNODO_DEFAULT_DOCKER_TAR_PATH}:/tmp/${SUNODO_DEFAULT_DOCKER_TAR_PATH}`,
```
tuler, line 192: I did not understand why we need this volume mapped, if the `docker-image.tar` does not need to be in the host filesystem

tuler:
We should have a single docker image that is used by the `build` command and the `run` command. This will avoid issues of having incompatible emulator versions between the two stages, and also in the future facilitate the support of replacing the emulator with a custom one for advanced use cases.

endersonmaia:
> We should have a single docker image that is used by the `build` command and the `run` command. This will avoid issues of having incompatible emulator versions between the two stages, and also in the future facilitate the support of replacing the emulator with a custom one for advanced use cases.

I plan to tackle this with #356 

What do you think is missing for us to close this PR without derailing into what I want to do on a separate PR?

tuler:
> What do you think is missing for us to close this PR without derailing into what I want to do on a separate PR?

Users who upgrade sunodo will also need to update their application Dockerfile to specify `sunodo/sdk:0.3.0` as the version of the sdk.
Those who don't do it will fail to build with an unpolite crash, because there is no `undocker` binary in the `sunodo/sdk:0.2.0` image.

One solution is to check the label and politely ask the user to update the Dockerfile

tuler (approved):

PR #377, opened 2024-02-21
Additional fixes of pnpm migration

tuler:
pnpm has a different way of specifying internal dependencies than yarn.
Some missing packages were also added, because yarn was masking this error.


endersonmaia (approved):
Some packages are still failing on the CI.

tuler:
> Some packages are still failing on the CI.

I know, but it's an improvement. Do you want to merge this, apply the fixup and we continue with the rest in the other?

Issue #386, opened 2024-02-25
Sunodo send ERC20 doesnt request amount


felipeargento:
## 🙂 Expected behavior

When using:
`sunodo send` to deposit ERC20, the cli should let me choose the amount to send, rather than sending an unspecified default amount.

## 🫠 Actual behavior

<img width="820" alt="image" src="https://github.com/sunodo/sunodo/assets/10563832/0dd65efc-8f6b-478e-86bc-00def728966b">

And it deposited 1:
`validator-1  | DBG received deposit deposit="0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 deposited 1000000000000000000 of 0xae7f61eCf06C65405560166b259C54031428A9C4 token"`
## 🧪 Minimal test case
```
sunodo create
sunodo build
sunodo run

# Send approval transaction for sunodo test token:
cast send 0xae7f61eCf06C65405560166b259C54031428A9C4 0x095ea7b30000000000000000000000009C21AEb2093C32DDbC53eEF24B873BDCd1aDa1DB00000000000000000000000000000000000000000000000ffffffffffff0ffff --rpc-url http://localhost:8545 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80

sunodo send ERC20
```

## 🌎 Environment

VERSION
  @sunodo/cli/0.10.4 darwin-arm64 node-v21.6.1

## ✔️ Possible solutions

Add a prompt for asking the amount, just like it asks for the dapp address etc

PR #389, opened 2024-02-27
Asking token amount when sending ERC20

DiegoSena:
Fixes https://github.com/sunodo/sunodo/issues/386.

make sunodo send erc20 ask for the amount

### DIFF
--- a/apps/cli/src/commands/send/erc20.ts
+++ b/apps/cli/src/commands/send/erc20.ts
@@ -26,7 +26,7 @@ export default class SendERC20 extends SendBaseCommand<typeof SendERC20> {
 
     static flags = {
         token: CustomFlags.address({ description: "token address" }),
-        amount: CustomFlags.number({ description: "amount", default: "1" }),
+        amount: CustomFlags.number({ description: "amount" }),
     };
 
     static examples = ["<%= config.bin %> <%= command.id %>"];
--- a/.changeset/hungry-moons-invent.md
+++ b/.changeset/hungry-moons-invent.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/cli": patch
+---
+
+make sunodo send erc20 ask for the amount


apps/cli/src/commands/send/erc20.ts
```diff
@@ -83,8 +83,7 @@ export default class SendERC20 extends SendBaseCommand<typeof SendERC20> {
                 validate: ercValidator,
             })) as Address);
 
-        const amount =
-            this.flags.amount || (await input({ message: "Amount" }));
+        const amount = await input({ message: "Amount", default: this.flags.amount });
```
tuler, line 86: Instead of removing the option ofreading the flag, the right way to fix this is remove the default value of the flag up there in line 29

DiegoSena: Fixed.

apps/cli/src/commands/send/erc20.ts
```diff
@@ -85,7 +85,7 @@ export default class SendERC20 extends SendBaseCommand<typeof SendERC20> {
 
         const amount =
             this.flags.amount || (await input({ message: "Amount" }));
-
+
```
tuler, line 88: Can you remove this change?

tuler:
You need to write a changeset and commit together.
Information on how to do it is above in this PR.

apps/cli/src/commands/send/erc20.ts
DiegoSena: Done

tuler:
Can you squash the commits to a single one with a message that has the format:
`fix(cli): fix sunodo send erc20 not asking for token amount`

DiegoSena:
> Can you squash the commits to a single one with a message that has the format: `fix(cli): fix sunodo send erc20 not asking for token amount`

Done

endersonmaia (approved):

Issue #200, opened 2023-09-11
Set allowance during ERC-20 deposit


tuler:
## 📚 Context

The ERC-20 contract requires to set an allowance before depositing tokens into the ERC20Portal contract.
Sunodo `send erc20` does not handle allowance, so it often fails like so:

```
? Select send sub-command Send ERC-20 deposit to the application.
? Chain Foundry
? RPC URL http://127.0.0.1:8545
? Wallet Mnemonic
? Mnemonic test test test test test test test test test test test junk
? Account 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 9999.969156636810215241 ETH
? DApp address 0x70ac08179605AF2D9e75782b8DEcDD3c22aA4D0C
? Token address 0xae7f61eCf06C65405560166b259C54031428A9C4
    ContractFunctionExecutionError: The contract function "depositERC20Tokens" reverted with the following reason:
    ERC20: insufficient allowance

    Contract Call:
      address:   0x9C21AEb2093C32DDbC53eEF24B873BDCd1aDa1DB
      function:  depositERC20Tokens(address _token, address _dapp, uint256 _amount, bytes _execLayerData)
      args:                        (0xae7f61eCf06C65405560166b259C54031428A9C4, 0x70ac08179605AF2D9e75782b8DEcDD3c22aA4D0C, 1000000000000000000, 0x)
      sender:    0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266

    Docs: https://viem.sh/docs/contract/simulateContract.html
    Version: viem@1.9.3
```

## ✔️ Solution

Sunodo `send erc20` could also take care of the allowance transaction.

## 🎯 Definition of Done

- [ ] `send erc20` asks for allowance if necessary, and send the allowance transaction before sending the ERC20Portal transaction.

tuler:
Moved to https://github.com/cartesi/cli/issues/22

PR #395, opened 2024-03-01
feat(cli):set allowance for deposit if necessary

DiegoSena:
Fixes https://github.com/sunodo/sunodo/issues/200

Set allowance for deposit if necessary


apps/cli/src/commands/send/erc20.ts
```diff
@@ -8,7 +8,7 @@ import {
     WalletClient,
 } from "viem";
 
-import { erc20PortalAbi, erc20PortalAddress } from "../../contracts.js";
+import { erc20PortalAbi, erc20PortalAddress, sunodoTokenAbi } from "../../contracts.js";
```
tuler, line 11: You don't need to use a `sunodoTokenAbi`. Use the `erc20Abi` from `viem`.
The feature is valid for any ERC-20 token, not SunodoToken.

apps/cli/src/commands/send/erc20.ts
```diff
@@ -85,6 +118,8 @@ export default class SendERC20 extends SendBaseCommand<typeof SendERC20> {
 
         const amount =
             this.flags.amount || (await input({ message: "Amount" }));
+
```
tuler, line 121: Parse the value in this point to bigint with `parseEther`, and remove the extra parsing inside setAllowance.

apps/cli/src/commands/send/erc20.ts
```diff
@@ -55,6 +55,39 @@ export default class SendERC20 extends SendBaseCommand<typeof SendERC20> {
         };
     }
 
+    private async setAllowance(
+            publicClient: PublicClient,
+            walletClient: WalletClient,
+            amount: string,
```
tuler, line 61: use bigint here, see comment below

apps/cli/src/commands/send/erc20.ts
```diff
@@ -55,6 +55,39 @@ export default class SendERC20 extends SendBaseCommand<typeof SendERC20> {
...
+            tokenAddress: Address) {
+
+        const allowance = await publicClient.readContract({
+            address: tokenAddress,
+            abi: sunodoTokenAbi,
```
tuler, line 66: this should be erc20Abi

apps/cli/src/commands/send/erc20.ts
```diff
@@ -55,6 +55,39 @@ export default class SendERC20 extends SendBaseCommand<typeof SendERC20> {
...
+            publicClient: PublicClient,
+            walletClient: WalletClient,
+            amount: string,
+            tokenAddress: Address) {
+
+        const allowance = await publicClient.readContract({
+            address: tokenAddress,
+            abi: sunodoTokenAbi,  
+            functionName: "allowance",
+            args: [walletClient.account?.address as `0x${string}`, erc20PortalAddress],
```
tuler, line 68: Instead of passing the walletClient as parameter, pass the owner address only.

DiegoSena: I still need the walletClient to writeContract and also need the account that's why i went for the object.

apps/cli/src/commands/send/erc20.ts
DiegoSena: Done

apps/cli/src/commands/send/erc20.ts
DiegoSena: done

apps/cli/src/commands/send/erc20.ts
DiegoSena: done

tuler (approved):
Good job!

tuler:
This is not working as expected.
In the first run it reverts. In the second run it works. In the third run it reverts.
It's inconsistent, not sure what is going on yet.

tuler:
Also please configure vscode to use prettier. Code is not formatted.

tuler:
One thing that for sure is missing is that you have to wait for the transaction to be mined before moving forward.
In the structure of the send commands this is done by the `SendBaseCommand` class. including a proper `ora` progress bar.
Sending an additional transaction like how you are doing, is not well aligned with the structure of the `SendBaseCommand`. We should rethink how to do it elegantly.

tuler:
Comment above still applies

Issue #390, opened 2024-02-29
Add a command that prints out the template hash


guidanoli:
## 📚 Context

Post-build, the image hash is stored in the `.sunodo/image/hash` file in binary format.
For most use cases, however, a hexadecimal format is more useful.
This hash can be displayed in this format by running the following command.

```sh
xxd -c 32 -p .sunodo/image/hash
```

This might not be very intuitive for beginner developers, though.

## ✔️ Solution

Add a new command, e.g. `sunodo hash`, which prints the machine hash in hexadecimal format (also, prefixed by `0x`). The `xxd` utility should not be necessary.

## 📈 Subtasks

-   [ ] Add a `sunodo hash` command
-   [ ] Mention the command in the documentation
-   [ ] Test the command against the traditional method
-   [ ] Test the command when no image has been built yet
-   [ ] Test the command when the image was built but the `hash` file is missing

## 🎯 Definition of Done

-   [ ] The command was added
-   [ ] The command was documented
-   [ ] The command was tested

PR #399, opened 2024-03-06
Hash command

DiegoSena:
Fixes https://github.com/sunodo/sunodo/issues/390

Adds hash command that prints outs the hash generated by build command


apps/cli/src/commands/hash.ts
```diff
@@ -0,0 +1,22 @@
...
+        const hashPath = path.join(".sunodo", "image", "hash");
+
+        if(fs.existsSync(hashPath)) {
+            const hash = fs.readFileSync(hashPath).toString("hex");
+
+            this.log(hash)
+            return;
+        }
+
+        this.log("Hash file not found")
```
tuler, line 19: In this case you should call `this.error`

apps/cli/src/commands/hash.ts
```diff
@@ -0,0 +1,22 @@
...
+import path from "path";
+
+
+export default class HashCommand extends Command {
+    
+    public async run() {
+        // read hash of the cartesi machine snapshot, if one exists
+        const hashPath = path.join(".sunodo", "image", "hash");
+
+        if(fs.existsSync(hashPath)) {
```
tuler, line 12: You may be missing prettier format here

apps/cli/src/commands/hash.ts
```diff
@@ -0,0 +1,22 @@
...
+            const hash = fs.readFileSync(hashPath).toString("hex");
+
+            this.log(hash)
```
tuler, line 15: We should also support json output. Take a look at https://oclif.io/docs/json
Do a `if (this.jsonEnabled())`.
It's similar to what it's done at the AddressBook command, but not the same (you don't need to use `ux.table` in this case.

apps/cli/src/commands/hash.ts
DiegoSena: done

apps/cli/src/commands/hash.ts
DiegoSena: done

README.md
```diff
@@ -12,6 +12,7 @@ It helps developers to:
 -   `deploy` applications to a cloud provider
 -   `monitor` application already running on a cloud provider
 -   `doctor` verify the minimal requirements for the Sunodo CLI commands
+-   `hash` prints out image hash generated by the build command
```
endersonmaia, line 15: It would be nice to be more specific that this hash is the `templateHash` of the snapshot, since the container image also has a hash and this could led to confusion.

Not sure about the best wording here, if `tempalteHash` or just `hash` or `snapshot hash`.

```suggestion
-   `hash` prints out Cartesi Machine tempalteHash generated by the build command
```

endersonmaia (commented):
On the commit message, let's use leading active verbs. Instead of `adding`, let's use `add`

My suggestion, in sync with the changeset:

```
feat(cli): add hash command
```

README.md
tuler: Agree, just you added a typo there

apps/cli/src/commands/hash.ts
tuler: Replace this line log with:
```
process.stdout.write(
        `${chalk.green("?")} Machine hash ${chalk.cyan(templateHash)}\n`,
    );
```

endersonmaia:
Will me merge at #403

PR #400, opened 2024-03-08
Packages missing rimraf for clean task

tuler:
After migration to pnpm, rimraf is missing from some packages.

ps: I don't think we need a changeset for this.


tuler:
> Are going to do anything about the Socket security check?

I'm not sure yet if the alerts are working correctly.

Issue #385, opened 2024-02-25
Address of test token reported by address-book is wrong


tuler:
## 🙂 Expected behavior

The address of the `SunodoToken` reported by `sunodo address-book` is `0xf795b3D15D47ac1c61BEf4Cc6469EBb2454C6a9b`, but the token is actually deployed at `0xae7f61eCf06C65405560166b259C54031428A9C4`.

## 🫠 Actual behavior

The devnet deployment is not matching somehow the address of the contracts.

## 🧪 Minimal test case

Use the wrong address to try to make a deposit.

## 🌎 Environment

sunodo 0.10.4

## ✔️ Possible solutions

This can be reviewed as part of issue https://github.com/sunodo/sunodo/issues/381

PR #308, opened 2023-12-12
Adapt to rollups node 1.3.0

endersonmaia:
This PR will release `@sunodo/cli:0.11.0` and `sunodo/rollups-node:0.7.0` ready to use the new  `cartesi/rollups-node:1.3.1` version.

Depends on: Final release https://github.com/cartesi/rollups-node/releases/tag/v1.3.1

PR #405, opened 2024-03-11
Devnet contracts

### LINKED ISSUE #381: Add instantiation of required contracts to sunodo/devnet
## 📚 Context

Rollups 1.0 used to have an Authority and a History contract already deployed and included in the npm package.
Those contracts are necessary to run a node.

Rollups 1.2 does not include those contract any longer. It's a responsibility of the node runner to deploy those contracts prior to running a node.

The devnet environment could do that for the user, and have a docker image with anvil ready to be used by a local development environment.

The upcoming marketplace feature will also need a ValidatorNodeProvider contract instantiated.

## ✔️ Solution

Write a hardhat task in the devnet project that instantiates the required contracts.
Call that task during build to produce the anvil dump we are baking into a docker anvil image.
We also need to store the address of the contracts in some way inside the image.

## 📈 Subtasks

- [x] #384
- [x] write a task to instantiate a ValidatorNodeProvider
- [x] modify build process to call the tasks

## 🎯 Definition of Done

- [x] sunodo/devnet image has information about instantiated contracts

tuler:
This creates a `Devnet` contract to the `devnet` project that instantiates an authority, a history and a provider, with simple defaults for a devnet environment.

The `pnpm run build --filter devnet` from root does the following:

1. build `contracts` project and `token` project, as they are dependencies to the `devnet` project.
2. compile solidity contract of devnet project
3. run an anvil instance with --dump-state (on exit)
4. deploy contracts to anvil through hardhat-deploy external script (cartesi contracts, sunodo contracts, sunodo token), exporting those to a json
5. deploy an instance of the `Devnet` contract
6. read addresses of authority, history, provider from `Devnet` instance, and patch the json exported in step 3 with these new contracts
7. shutdown anvil, writing dump state
8. build a docker image from the `anvil` image with the deploy files and the dumped state.

closes #381, #385


tuler:
This will also fix #385

packages/devnet/contracts/Devnet.sol
```diff
@@ -0,0 +1,41 @@
...
+import {IConsensus} from "@cartesi/rollups/contracts/consensus/IConsensus.sol";
+import {Authority} from "@cartesi/rollups/contracts/consensus/authority/Authority.sol";
+import {History} from "@cartesi/rollups/contracts/history/History.sol";
+import {IAuthorityHistoryPairFactory} from "@cartesi/rollups/contracts/consensus/authority/IAuthorityHistoryPairFactory.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {IMarketplace} from "@sunodo/contracts/contracts/marketplace/IMarketplace.sol";
+import {IValidatorNodeProvider} from "@sunodo/contracts/contracts/provider/INodeProvider.sol";
+
+contract Devnet {
+    IERC20 public immutable token;
```
guidanoli, line 14: Is the `token()` function ever called, once the `Devnet` contract is deployed?
If not, we can remove this state variable and the assignment at line 24.

tuler: Removed

tuler:
I'm rethinking if we need this deployment of `DevnetAuthority` and `DevnetHistory` considering the introduction of the `SelfHostedApplicationFactory` of PR https://github.com/sunodo/sunodo/pull/409

We can just call the `SelfHostedApplicationFactory` in the `dapp_deployer` container, and hardcode all addresses (authority, history, dapp) since we can pre-calculate it.

We can use a zero hash as the templateHash, so we have a fixed dapp address, and disable the hash validation of the node using the option CARTESI_FEATURE_DISABLE_MACHINE_HASH_CHECK

endersonmaia:
I suggest this PR should be on top of #308.

guidanoli:
Shouldn't this PR have `feature/adapt-to-rollups-node-1.3.0` as the base branch?

tuler:
I'll close this PR in favor of #409 
With deterministic deployment we don't need the `Devnet` contract deployment, because we can hard-code contract addresses.

Issue #387, opened 2024-02-26
`sunodo create` should allow other git repository as source


endersonmaia:
## 📚 Context

Currently, the `sunodo create` command allow to define a `--branch` to use as source of a template, but that's mainly an internal sunodo developer option.

So if a regular user needs to use another git repository other than sunodo/sunodo-template it's now possible.

## ✔️ Solution

Add a `--repository` argument where the use can define the repository to use as the source of template.

### Requirements

- `main` should be the default branch when using this option
- the template name should be used as the directory name `./$template-name` to get its content

## 📈 Subtasks

-   [ ] Add a `--repository` argument to `create` command following the previous **Solution** section

## 🎯 Definition of Done

-   [ ] A user can define a different git repository as source to create a local application from template

DiegoSena:
@endersonmaia What do you think it should happen if the repository/branch doesn't exist ? Fail with a message ?

tuler:
Another solution that can be much more powerful and flexible, and not so difficult to implement is to use yeoman as the template engine for scaffolding.
Yeoman is a mature project that has a CLI tool (`yo`), but can also be used as a library and integrated into `sunodo create`.
https://yeoman.io/authoring/integrating-yeoman

Issue #186, opened 2023-09-01
sunodo create should fail when --branch doesn't exists


endersonmaia:
## 🙂 Expected behavior

When calling `sunodo create --branch=wrong-branch-name`, the CLI should show an error informing the branch doesn't exists.

## 🫠 Actual behavior

The CLI show that the application was successfully created, instead.

## 🧪 Minimal test case

```
$ sunodo create --template=javascript --branch=wrong-branch-name my-js-dapp
✔ Application created at /home/endersonmaia/Development/my-js-dapp
$ ls -1 my-js-dapp/
Dockerfile
package.json
README.md
src
yarn.lock
```

## 🌎 Environment

What is the environment that the bug happens?

I'm using sunodo CLI 0.7.0

DiegoSena:
@tuler All braches on sunodo-templates are eligible or do we have a fixed list ?

endersonmaia:
@DiegoSena we lock the branch used based on sunodo/sdk version.

For ex., for @sunodo/cli:0.10.4, we use the branch `sdk-0.2` as you can see here https://github.com/sunodo/sunodo/blob/%40sunodo/cli%400.10.4/apps/cli/src/commands/create.ts#L7

Is that what you asked?

DiegoSena:
Not exactly. We can use branch flag on CLI and can literally use any branch on sunodo-template repo. We should allow select ANY BRANCH or based on sunodo version we can select only a few listed branches?

tuler:
I think that flag is more targeted to developers who want to test new templates. Regular users should use just the “main” branch configured for the CLI version.

Maybe the flag could be hidden?

endersonmaia:
That and there's an idea to make it possible to define a different repository also.

In that case, it makes more sense for regular users. Maybe that could go on another issue.

PR #407, opened 2024-03-11
feat(cli): Adds repository flag

DiegoSena:
Fixes https://github.com/sunodo/sunodo/issues/387

Adds repository flag to create command


apps/cli/src/commands/create.ts
```diff
@@ -34,23 +35,23 @@ export default class CreateCommand extends Command {
...
+        repository: string
     ): Promise<DownloadTemplateResult> {
         const sunodoProvider: TemplateProvider = async (input) => {
             return {
                 name: "sunodo",
                 subdir: input,
-                url: "https://github.com/sunodo/sunodo-templates",
-                tar: `https://codeload.github.com/sunodo/sunodo-templates/tar.gz/refs/heads/${branch}`,
+                url: repository,
+                tar: this.getTarUrl(repository),
```
endersonmaia, line 54: Keep the `sunodoProvider`  name here won't  make sense, maybe a more generic name like `repositoryProvider` or `githubProvider` since it may only work with github.

apps/cli/src/commands/create.ts
```diff
@@ -34,23 +35,23 @@ export default class CreateCommand extends Command {
                 "typescript",
             ],
         }),
-        branch: Flags.string({
-            description: `sunodo/sunodo-templates repository branch name to use`,
-            default: DEFAULT_TEMPLATES_BRANCH,
-        }),
+        repository: Flags.string({
```
endersonmaia, line 38: There's no reason to stop supporting the `--branch` flag.

There's an argument for hiding it here https://github.com/sunodo/sunodo/issues/186#issuecomment-1962992079

apps/cli/src/commands/create.ts
```diff
@@ -61,14 +62,26 @@ export default class CreateCommand extends Command {
...
+        return repository.replace("github.com", "codeload.github.com");
+    }
+
+    private getTarUrl(repository: string) {
+        if (repository === DEFAULT_TEMPLATE_REPOSITORY) {
+            return `${this.getRepoCodeLoadUrl(repository)}/tar.gz/refs/heads/${DEFAULT_TEMPLATES_BRANCH}`;
+        }
+        return `${this.getRepoCodeLoadUrl(repository)}/tar.gz/refs/heads/main`;
+    }
+
```
endersonmaia, line 75: This will be too error prone, since it will only work for github.

It looks like giget support a lot of other providers.

https://github.com/unjs/giget?tab=readme-ov-file#downloadtemplatesource-options

endersonmaia (changes requested):

tuler:
Back to the drawing board on this one

PR #413, opened 2024-03-15
Deprecate sunodo rollups node

### LINKED ISSUE #411: Deprecate sunodo/rollups-node in favor of using cartesi/rollups-node directly
## 📚 Context

We have a sunodo/rollups-node because we added a lot of other tools to make cartesi/rollups-node run in a single container and other stuff.

Since the release of cartesi/rollups-node:1.3.0, with the go-supervisor functionality, we don't need to manage the internal services anymore, and the sunodo/rollups-node has only the rollups-node service and a snapshot-provisioner script.

## ✔️ Solution

We should use cartesi/rollups-node container image directly from sunodo/cli and deprecate de usage of sunodo/rollups-node.

## 📈 Subtasks

-   [x] Change sunodo/cli to use cartesi/rollups-node container image
-   [x] Make snapshot provisioning

## 🎯 Definition of Done

-   [x] When sunodo/cli doesn't depends on sunodo/rollups-node anymore

endersonmaia:
This PR will remove the @sunodo/rollups-node package and make @sunodo/cli use the cartesi/rollups-node container image instead of using sunodo/rollups-node image.

Close #411


apps/cli/src/node/docker-compose-validator.yaml
```diff
@@ -2,7 +2,8 @@ version: "3.9"
 
 services:
     validator:
-        image: sunodo/rollups-node:0.7.0
+        image: cartesi/rollups-node:1.3.1
+        user: root
```
tuler, line 6: do we need to force root here?

endersonmaia: Cartesi/rollups-node uses the cartesi user

We need root to make the snapshot copy

apps/cli/src/commands/run.ts
```diff
@@ -72,12 +72,12 @@ export default class Run extends Command {
         const epochDuration = flags["epoch-duration"];
         const listenPort = flags["listen-port"];
         const env: NodeJS.ProcessEnv = {
+            CARTESI_SNAPSHOT_DIR: "/usr/share/cartesi/snapshot",
```
tuler, line 75: let's try to sort this list in alphabetical order

endersonmaia: sent a fixup

endersonmaia:
If we can have a cartesi/rollups-node:1.3.2 patch release https://github.com/cartesi/rollups-node/pull/356 we may add a fixup here so we don't need to run as `root` just to copy the snapshot.

endersonmaia:
> If we can have a cartesi/rollups-node:1.3.2 patch release [cartesi/rollups-node#356](https://github.com/cartesi/rollups-node/pull/356) we may add a fixup here so we don't need to run as `root` just to copy the snapshot.

This won't be necessary, since the directory is OK, we need to fix the mounting where the snapshot is provided, to avoid `cp: cannot stat '/tmp/snapshot/*': Permission denied` error.

The directory has this permissions:

```
$ ls -lgha /tmp/
total 12K
drwxrwxrwt 1 root 4.0K Mar 15 17:09 .
drwxr-xr-x 1 root 4.0K Mar 15 17:09 ..
drwx------ 2 1000 4.0K Mar 12 11:34 snapshot
```

tuler (approved):

PR #417, opened 2024-03-19
feat(cli): add --verbose flag to sunodo build command

endersonmaia:
This PR will add a `--verbose` flag to `sunodo build`, making the normal mode of `sunodo build` more quiet.

### DIFF
--- a/.changeset/fair-flies-lay.md
+++ b/.changeset/fair-flies-lay.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/cli": patch
+---
+
+add --verbose flag to sunodo build command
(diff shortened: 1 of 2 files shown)


tuler:
I didn't like this very much because in case of docker build error, which is very common, the error is not displayed.
I think we will be safer if we leave the docker build as it is.

On a less important note, we also lose docker color output during `--verbose`

endersonmaia:
> On a less important note, we also lose docker color output during `--verbose`

That's because I used `--progress=plain`, we can just don't change this.

tuler:
Let the docker pull print

PR #421, opened 2024-03-21
Refactor/sunodo build simpler execa calls

endersonmaia:


tuler:
What I don’t like in that direction is that IMO this increases maintenance effort instead of reducing, because it adds up another layer of variables names (environment variables in this case) that we need to take care to update.

tuler:
It would be different if the binaries we use supported env variables, then it would just be a passthrough of the variables.

endersonmaia:
Still a draft...

I have another PR where I try to wrap the calls inside a generic command API, that could be another way of doing that

PR #434, opened 2024-04-04
Remove mobile wallets

tuler:
Remove mobile wallets support.
Main intent is to reduce package size, in exchange of removing two features that I think are not being used.

Also removes some unused packages.

Reduced total package size from ~280Mb to ~48Mb.


endersonmaia (changes requested):
LGTM, just need to rebase and fix conflicts, so I can approve.

endersonmaia (approved):

PR #435, opened 2024-04-04
Migrate from oclif to clipanion

tuler:
Replaces oclif with clipanion.

I'm still unsure if this is the best alternative, or if `commander` is better.

Commander is far more popular, but clipanion is more elegant, which probably reflected on how easy it was to migrate.

I prefer commander help than clipanion help format.

Clipanion maintenance is a bit unclear.

Reduces package size even further, from 48Mb to 21.7Mb

closes https://github.com/sunodo/sunodo/issues/428


endersonmaia:
Please, consider moving this do `cartesi/cli` and rebasing.

Issue #404, opened 2024-03-11
Migrate from Sunodo CLI to Cartesi CLI


endersonmaia:
This issue is meant to group all the effort needed to migrate to the future https://github.com/cartesi/cli project.

- [x] #447
- [x] #445
- [x] https://github.com/sunodo/sunodo/pull/493
- [ ] ~~Define owner and repository to hold cartesi marketplace smart contracts~~
- [x] Create `cartesi/application-templates` for `cartesi create` command
- [x] Publish @cartesi/cli@0.14.0
- [x] Publish brew formula at https://github.com/cartesi/homebrew-tap
- [x] https://github.com/cartesi/docs/pull/183
- [x] https://github.com/sunodo/sunodo/pull/492
- [x] https://github.com/sunodo/sunodo/pull/495

For context: 

- https://governance.cartesi.io/t/sunodo-a-cartesi-convenience-layer/370/2?u=endersonmaia
- https://twitter.com/SunodoProject/status/1759578710599241986

tuler:
Consider a version where the transition will happen. I'll call it here 0.13, but it may change.
On the user's perspective, this is what would happen.

User upgrades sunodo to 0.13.
Any command he makes (sunodo create, sunodo build, etc) prints the following message:

```
DEPRECATED: Sunodo CLI is now Cartesi CLI.
Uninstall Sunodo CLI by following the documentation at https://docs.sunodo.io/guide/introduction/installing;
Install Cartesi CLI by following the documentation at https://docs.cartesi.io/cartesi-rollups/1.3/development/installation/
```

Cartesi CLI publishes a 0.14 version that has everything pre-sunodo 0.13.

Here is how we can make the code transition.
Cartesi forks the Sunodo repo. On top of main@0.13 it reverts the changes that removed the commands functionality (which were replaced to add the DEPRECATION message). It adds changes to whatever it needs to publish the packages as `@cartesi`. And publishes a `@cartesi/cli@0.14`.

endersonmaia:
> Cartesi forks the Sunodo repo

I suggest that we transfer so we don't lose issues and stuff.

https://github.com/sunodo/sunodo/issues/447#issue-2233080766

tuler:
## These are the current packages and their internal dependencies:

```
├── apps
│   ├── cli
│   ├── docs
│   └── web
└── packages
    ├── car-download
    ├── contracts
    ├── devnet
    ├── eslint-config
    ├── sdk
    └── tsconfig
```

[...]

PR #448, opened 2024-04-09
Feat/npm sunodo release deprecation warning

endersonmaia:


tuler:
I don't think we need a warning now. When next version comes the user will need to uninstall sunodo and install cartesi.
See considerations at #404

PR #459, opened 2024-04-12
Refactor build command

endersonmaia:
This PR will introduce a `sdkRun()` internal function that will work as a one-shot call to executing command using the `sunodo/sdk` image, and processing an `input` and saving an `output`.

The idea is that we don't need to have a running container to run the rootfs -> ext2 -> snapshot "pipeline".

Some things are not elegant TypeScript code, but the overall idea is there.

### DIFF
--- a/apps/cli/src/commands/build.ts
+++ b/apps/cli/src/commands/build.ts
@@ -145,111 +145,93 @@ Update your application Dockerfile using one of the templates at https://github.
         return info;
     }
 
-    private async startBuildContainer(
-        appImage: string,
-        builderImage: string,
-    ): Promise<string> {
+    // creates a rootfs tarball from the image
+    // this process is not always fully reproducible
+    // FIXME: we could use the image and create a flat rootfs without
+    //        `docker container create` (use undocker, umoci, a native typescript implementation, etc.)
+    private async createTarball(
+        image: string,
+        outputFilePath: string,
+    ): Promise<void> {
         // create docker tarball from app image
         const { stdout: appCid } = await execa("docker", [
             "container",
             "create",
             "--platform",
             "linux/riscv64",
-            appImage,
+            image,
         ]);
 
         await execa("docker", [
             "container",
             "export",
             "-o",
-            SUNODO_DEFAULT_TAR_PATH,
+            outputFilePath,
             appCid,
         ]);
 
-        // start build container
+        await execa("docker", ["container", "rm", appCid]);
+    }
+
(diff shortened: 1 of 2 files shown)


apps/cli/src/commands/build.ts
```diff
@@ -145,111 +145,94 @@ Update your application Dockerfile using one of the templates at https://github.
...
+    // this wraps the call to the sdk image with a one-shot approach
+    // the (inputPath, outputPath) signature will mount the input as a volume and copy the output with docker cp
+    private async sdkRun(
+        sdkImage: string,
+        cmd: string[],
+        inputPath: string,
+        outputPath: string,
+    ): Promise<void> {
+        const randomPrefix = Math.random().toString(36).substring(7);
+        const containerName = `sdk-run-${randomPrefix}`;
```
tuler, line 185: You don’t need a random container name. Get the container id as we do above

endersonmaia: but we get the container id from the `docker container create` command

To get the container ID from `docker container run` I'd have to use `--cidfile=` and read from the file, seems more work than just defining a random value at runtime. What do you think?

endersonmaia: sent a fixup

endersonmaia (commented):
I'm getting this erro when trying to send a bounty via cli.

```shell
❯ go run ./cli send bounty \
    -n "Lua Bounty" \
    -d "Description of Lua bounty" \
    -c ./tests/bounties/lua-bounty/lua-5.4.3-bounty_riscv64.tar.xz
INF fork/exec /usr/local/bin/cast: argument list too long
exit status 1
```

tuler:
I don't like very much the style of the resulting code when we have methods like `createRootfsTarCommand` and `createExt2Command` which don't receive any argument, and just return a `string[]`, making it pointless to be a method.

It didn't need to be a class method (could be a static one), it didn't need to be async (it has no async call), and in the end it didn't even need to be a method.

endersonmaia:
> I don't like very much the style of the resulting code where we have methods like `createRootfsTarCommand` and `createExt2Command` which don't receive any argument, and just return a `string[]`, making it pointless to be a method.
> 
> It didn't need to be a class method (could be a static one), it didn't need to be async (it has no async call), and in the end it didn't even need to be a method.

Yeah, I felt the same, but didn't want to change since it was open to review.

Any suggestion that I can fix quickly?

PR #460, opened 2024-04-12
Support to base network

tuler:
This completes support to base network.
It adds base to the deploy web app.
It adds base to CLI in wallet selection for sunodo send.
It adds base to documentation.

### DIFF
--- a/.changeset/long-rats-mate.md
+++ b/.changeset/long-rats-mate.md
@@ -1,0 +1,7 @@
+---
+"@sunodo/docs": patch
+"@sunodo/cli": patch
+"@sunodo/web": patch
+---
+
+support to base network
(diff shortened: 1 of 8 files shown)


endersonmaia (approved):
Isn´t there a base-sepolia network?

tuler:
> Isn´t there a base-sepolia network?

I wasn't aware until yesterday there is a base-sepolia network.

If we really want to deploy to base-sepolia we need to add it all the way from the `solidity-util` package, to the `rollups-contracts` package to here.

Do you think it's necessary?

tuler:
> Do you think it's necessary?

For consistency it's probably better to do it.

tuler:
But I think we can merge this one, and then add base sepolia

PR #464, opened 2024-04-15
fix: check docker and docker compose in sunodo doctor

### LINKED ISSUE #412: sunodo doctor fails if docker compose is not installed
## 🙂 Expected behavior

When running `sunodo doctor` and `docker compose ` is not installed, the user should be informed that `docker compose` is a requirement and it should inform how to install it or links to instructions on how to do it.

## 🫠 Actual behavior

If the user doesn't have `docker compose` installed, `sunodo doctor` fails with an error.

```
Error: Command failed with exit code 125: docker compose version --short
unknown flag: --short
```

## 🧪 Minimal test case

Run `sunodo doctor` without `docker compose` installed and the error should show the error above.

## ✔️ Possible solutions

First check if `docker compose` in installed before checking for the version, and show pretty message informing the situation instead of an error.

omidasadpour:
/close #412 

If the user doesn't have docker : 

image

If the user doesn't have docker compose : 

image

Minimum docker version error : 

image

Minimum docker compose version error : 

image

### DIFF
--- a/apps/cli/src/commands/doctor.ts
+++ b/apps/cli/src/commands/doctor.ts
@@ -10,75 +10,131 @@ export default class DoctorCommand extends Command {
 
     private static MINIMUM_DOCKER_VERSION = "23.0.0"; // Replace with our minimum required Docker version
     private static MINIMUM_DOCKER_COMPOSE_VERSION = "2.21.0"; // Replace with our minimum required Docker Compose version
+    private static MINIMUM_BUILDX_VERSION = "0.13.0"; // Replace with our minimum required Buildx version
 
-    private static isDockerVersionValid(version: string): boolean {
-        return semver.gte(version, DoctorCommand.MINIMUM_DOCKER_VERSION);
-    }
-
-    private static isDockerComposeVersionValid(version: string): boolean {
-        const v = semver.coerce(version);
-        return (
-            v !== null &&
-            semver.gte(v, DoctorCommand.MINIMUM_DOCKER_COMPOSE_VERSION)
-        );
-    }
-
-    private static isBuildxRiscv64Supported(buildxOutput: string): boolean {
-        return buildxOutput.includes("riscv64");
-    }
-
-    public async run() {
+    private async checkDocker(): Promise<true | never> {
         try {
-            // Check Docker version
-            const { stdout: dockerVersionOutput } = await execa("docker", [
+            const { stdout: dockerVersion } = await execa("docker", [
                 "version",
                 "--format",
                 "{{json .Client.Version}}",
             ]);
-            const dockerVersion = JSON.parse(dockerVersionOutput);
 
-            if (!DoctorCommand.isDockerVersionValid(dockerVersion)) {
-                throw new Error(
-                    `Unsupported Docker version. Minimum required version is ${DoctorCommand.MINIMUM_DOCKER_VERSION}. Installed version is ${dockerVersion}.`,
-                );
-            }
-
-            // Check Docker Compose version
-            const { stdout: dockerComposeVersionOutput } = await execa(
-                "docker",
-                ["compose", "version", "--short"],
-            );
-            const dockerComposeVersion = dockerComposeVersionOutput.trim();
-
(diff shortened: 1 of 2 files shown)


apps/cli/src/commands/doctor.ts
```diff
@@ -44,18 +44,39 @@ export default class DoctorCommand extends Command {
...
+                    throw new Error(
+                        `Unsupported Docker Compose version. Minimum required version is ${DoctorCommand.MINIMUM_DOCKER_COMPOSE_VERSION}. Installed version is ${dockerComposeVersion}.`,
+                    );
+                }
+
+                // If execution reaches this point, Docker Compose is installed and its version is valid
+            } catch (error: any) {
+                // Specify the type of error as `any` or `unknown` or `Error`
+                // Check if the error indicates that Docker Compose is not recognized as a command
+                if (error.message.includes("Command failed")) {
```
endersonmaia, line 71: I think this is too error prone. In the `try` block you run other commands.

Fox ex., if I try to run `sunodo doctor` without `docker` installed, I get a weird error message.

```shell
❯ sunodo-dev doctor
 ›   Error: Command failed with ENOENT: docker version --format {{json .Client.Version}}
 ›   spawn docker ENOENT
```

I think it's better to isolate in different blocks.

- Check if docker in installed
- Check if docker has the >= required version
- Check is compose is isntalled
- Check is compose is the >= required version

endersonmaia: And maybe using the error code instead of relying on the error message content.

endersonmaia (changes requested):

apps/cli/src/commands/doctor.ts
```diff
@@ -29,6 +29,9 @@ export default class DoctorCommand extends Command {
 
     public async run() {
         try {
+            // Check if Docker is installed
+            await execa("docker", ["--version"]);
+
             // Check Docker version
             const { stdout: dockerVersionOutput } = await execa("docker", [
```
endersonmaia, line 36: `docker` is being called two times here.

We could call it once and use its output to check the version, catch an error and see `error.exitCode` to treat it correctly. `127 exitCode` would be a command not found, for example.

apps/cli/src/commands/doctor.ts
```diff
@@ -43,13 +46,18 @@ export default class DoctorCommand extends Command {
...
             }
 
+            // Check if Docker Compose is installed
+            await execa("docker", ["compose", "version"]);
+
             // Check Docker Compose version
             const { stdout: dockerComposeVersionOutput } = await execa(
                 "docker",
                 ["compose", "version", "--short"],
             );
```
endersonmaia, line 56: The same is happening here.

apps/cli/src/commands/doctor.ts
```diff
@@ -75,9 +83,21 @@ export default class DoctorCommand extends Command {
...
+                    } else if (error.message.includes("docker compose")) {
+                        this.error(
+                            "Docker Compose is required but not installed or the command execution failed. Please refer to the Docker Compose documentation for installation instructions: https://docs.docker.com/compose/install/",
+                        );
+                    } else {
+                        this.error(error.message);
+                    }
+                } else {
+                    this.error(String(error));
+                }
```
endersonmaia, line 100: You're relying on error messages when you should rely on exit codes.

I suggest you have individual `try...catch` blocks for each command being checked.

endersonmaia (changes requested):

apps/cli/src/commands/doctor.ts
omidasadpour: > You're relying on error messages when you should rely on exit codes.

Property 'code' does not exist on type 'Error'.ts(2339)

omidasadpour: > I suggest you have individual `try...catch` blocks for each command being checked.

Sure

endersonmaia:
I'll move this PR to cartesi/cli project.

tuler:
Moved to Cartesi CLI

Issue #367, opened 2024-02-20
Sunodo CLI dependency on Docker


tuler:
## 📚 Context

Docker is a complex software that brings a lot of value, but also a lot of problems to end-users. Different installation methods have different nuances. We like to recommend Docker Desktop to all users, but some of them do not use for one reason or another.

The Sunodo CLI uses Docker for two different things:

1) help build the application root filesystem;
2) execute binaries during application build and node execution that are not available in the host system

## ✔️ Solution

To reduce or completelly eliminate the dependency on Docker, the two use cases above must be addressed.
The issues below follow them independently.

## 📈 Subtasks

- [ ] https://github.com/sunodo/sunodo/issues/366
- [x] https://github.com/sunodo/sunodo/issues/332

## 🎯 Definition of Done

- [ ] not depend on Docker for Sunodo CLI

endersonmaia:
We also use docker and compose for `sunodo run` to start dependencies like traefik, anvil, subsquid, PostgreSQL.

tuler:
https://github.com/cartesi/cli/issues/70

PR #465, opened 2024-04-16
feat: add repository flag to create sub-command

omidasadpour:


apps/cli/src/commands/create.ts
```diff
@@ -36,28 +35,52 @@ export default class CreateCommand extends Command {
...
-                name: "sunodo",
-                subdir: input,
-                url: "https://github.com/sunodo/sunodo-templates",
-                tar: `https://codeload.github.com/sunodo/sunodo-templates/tar.gz/refs/heads/${branch}`,
-            };
-        };
+        let source;
+
+        if (repository) {
+            if (repository.includes("github.com")) {
```
tuler, line 54: I think that if we will continue to use the [gitget](https://www.npmjs.com/package/giget#examples) package, which is a little bloated, we should let it do the work, and don't try to do this extra processing here, because it adds maintenance.

endersonmaia:
I tested with a gitlab.com repository and it failed.

```shell
❯ sunodo-dev create --template=javascript --repository=https://gitlab.com/endersommaia/sunodo-templates --branch=main js-gl
✖ Error creating application: Failed to download https://gitlab.com/endersommaia/sunodo-templates/-/archive/main.tar.gz: 403 Forbidden
```

I checked that the source is available at `https://gitlab.com/endersonmaia/sunodo-templates/-/archive/main/sunodo-templates-main.tar.gz` instead.

I don´t know where would be the place to fix, if our code or giget's code.

apps/cli/src/commands/create.ts
endersonmaia: There's a "crazy" approach I used at #345, and we get rid of giget in the process.

endersonmaia:
> > I tested with a gitlab.com repository and it failed.
> > ```shell
> > ❯ sunodo-dev create --template=javascript --repository=https://gitlab.com/endersommaia/sunodo-templates --branch=main js-gl
> > ✖ Error creating application: Failed to download https://gitlab.com/endersommaia/sunodo-templates/-/archive/main.tar.gz: 403 Forbidden
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I checked that the source is available at `https://gitlab.com/endersonmaia/sunodo-templates/-/archive/main/sunodo-templates-main.tar.gz` instead.
> > I don´t know where would be the place to fix, if our code or giget's code.
> 
> your repo should be public

It is public.

omidasadpour:
> ```shell
> ❯ sunodo-dev create --template=javascript --repository=https://gitlab.com/endersommaia/sunodo-templates --branch=main js-gl
> ```

The repo URL U have set is not correct.

apps/cli/src/commands/create.ts
tuler: That just increase our [dependency on Docker](https://github.com/sunodo/sunodo/issues/367)

endersonmaia:
> > ```shell
> > ❯ sunodo-dev create --template=javascript --repository=https://gitlab.com/endersommaia/sunodo-templates --branch=main js-gl
> > ```
> 
> The repo URL U have set is not correct.

Yeah, anyway the final URL is looking at the wrong path.

apps/cli/src/commands/create.ts
endersonmaia: That's the reason for the quotes on "crazy". :)

omidasadpour:
> Yeah, anyway the final URL is looking at the wrong path.

Fixed ! 

```shell
sunodo create test --template python --repository https://gitlab.com/endersonmaia/sunodo-templates --branch main
```

image

apps/cli/src/commands/create.ts
omidasadpour: > I think that if we will continue to use the [gitget](https://www.npmjs.com/package/giget#examples) package, which is a little bloated, we should let it do the work, and don't try to do this extra processing here, because it adds maintenance.

Need to convert repository URLs to "github:owner/repo" or "gitlab:owner/repo" and ... format for the `downloadTemplate` function. Currently, directly using `downloadTemplate`, requiring manual URL conversion.

Example:

```code
const { source, dir } = await downloadTemplate("github:unjs/template");
```

omidasadpour: We have two options: 

1 - get the full repo address from user and then convert it to the appropriate format (Like what I did)
2 - add a flag to define the SCM type. (Like: `--repo-type` )
       example:
       
       ```shell
       sunodo create test --template python --repository omidasadpour/sunodo-templates  --branch main --repo-type github
       ```

tuler: I mean documenting the format that giget is expecting (i.e. gh:org/repo).
In any case the default should still be download from sunodo templates.

tuler:
We need to discuss how or if we will do this. Closing for now.

PR #466, opened 2024-04-17
feat(sdk): add crane

endersonmaia:
This PR will introduce the crane utility inside the sunodo/sdk package. 

With it, we don't need to create the rootfs tarball via docker create and docker export, but use the docker image save output directly to a rootfs tarball.

### DIFF
--- a/.changeset/tricky-cars-invite.md
+++ b/.changeset/tricky-cars-invite.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/sdk": patch
+---
+
+add crane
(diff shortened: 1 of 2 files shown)


.changeset/tricky-cars-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sunodo/sdk": patch
```
tuler, line 2: Should be minor?

endersonmaia: I'm not sure. 

It doesn't breaks user but adds an option functionality.

PR #470, opened 2024-04-18
Add shellpath arg to sunodo shell

### LINKED ISSUE #469: sunodo shell should provide an option to define the shell to use
## 📚 Context

The shell that's called by `sunodo shell` is hardcoded to `/bin/bash`.

It should be possible to define another shell.

## ✔️ Solution

Default to `/bin/bash` and add a `--shell` argument to define a different one.

`sundo shell  --shell=/bin/sh`

omidasadpour:
/close #469 

default shell is `/bin/bash`: 

image

could define the specific shell :
```sh
sunodo shell shellpath /bin/sh
```

image

### DIFF
--- a/.changeset/rude-seas-protect.md
+++ b/.changeset/rude-seas-protect.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/cli": minor
+---
+
+add shellpath arg to sunodo shell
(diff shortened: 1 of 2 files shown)


tuler:
See my comment at https://github.com/sunodo/sunodo/issues/469#issuecomment-2069734341

tuler:
Rediscuss the solution for #469, if any

PR #474, opened 2024-04-23
fixup! fix: docker and docker compose error in doctor sub-command

endersonmaia:


apps/cli/src/commands/doctor.ts
```diff
@@ -11,102 +11,104 @@ export default class DoctorCommand extends Command {
...
-            "version",
-            "--short",
-        ]);
-        return stdout;
-    }
-
-    private static isDockerVersionValid(version: string): boolean {
-        return semver.gte(version, DoctorCommand.MINIMUM_DOCKER_VERSION);
-    }
+    private async checkDocker(): Promise<true | never> {
```
endersonmaia, line 14: @tuler I'd like your opinion on this return signature `Promise<true | neve>`.

My goal was to make proper error treatment, but I see the `true` return is a little weird, since I neve use it.

apps/cli/src/commands/doctor.ts
```diff
@@ -11,102 +11,104 @@ export default class DoctorCommand extends Command {
...
+        return true;
+    }
+
+    private async checkBuildx(): Promise<true | never> {
+        try {
             const { stdout: buildxOutput } = await execa("docker", [
                 "buildx",
                 "ls",
+                "--format",
+                "{{.Platforms}}",
```
omidasadpour, line 78: I have already installed the `buildx` but `doctor` generates this error : 

```
    Error: Docker Buildx is required but not installed. Please refer to the Docker Desktop documentation for installation instructions: https://docs.docker.com/desktop/

```

```shell
docker buildx ls
```

Output: 

```
NAME/NODE       DRIVER/ENDPOINT STATUS  BUILDKIT PLATFORMS
default         docker                           
  default       default         running v0.13.1  linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
desktop-linux * docker                           
  desktop-linux desktop-linux   running v0.12.5  linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6

```

apps/cli/src/commands/doctor.ts
```diff
@@ -11,102 +11,104 @@ export default class DoctorCommand extends Command {
...
-                }
+            const v = semver.coerce(dockerComposeVersion);
+            if (
+                v !== null &&
+                !semver.gte(v, DoctorCommand.MINIMUM_DOCKER_COMPOSE_VERSION)
+            ) {
+                throw new Error(
+                    `Unsupported Docker Compose version. Minimum required version is ${DoctorCommand.MINIMUM_DOCKER_COMPOSE_VERSION}. Installed version is ${v}.`,
+                );
             }
```
omidasadpour, line 58: When I run doctor on a system without docker, It can't catch the error and show the message properly: 

```
    Error: Command failed with ENOENT: docker compose version --short
    spawn docker ENOENT
    Code: ENOENT
```

apps/cli/src/commands/doctor.ts
omidasadpour: `docker  buildx ls --format {{.Platforms}}` output: 

```
unknown flag: --format
See 'docker buildx ls --help'.
```

My version : 
```github.com/docker/buildx v0.12.1-desktop.4 6996841df2f61988c2794d84d33205368f96c317```

File: apps/cli/src/commands/doctor.ts (hunk 3 above)
Comment (endersonmaia) [reply apps/cli/src/commands/doctor.ts:58]: hmm, so we should only check for compose is docker is installed first

```patch
diff --git a/apps/cli/src/commands/doctor.ts b/apps/cli/src/commands/doctor.ts
index d37a624c..9498475b 100644
--- a/apps/cli/src/commands/doctor.ts
+++ b/apps/cli/src/commands/doctor.ts
@@ -102,9 +102,10 @@ export default class DoctorCommand extends Command {
 
     public async run(): Promise<void> {
         try {
-            await this.checkDocker();
-            await this.checkCompose();
-            await this.checkBuildx();
+            if (await this.checkDocker()) {
+                await this.checkCompose();
+                await this.checkBuildx();
+            }
         } catch (e: unknown) {
             this.error(e as Error);
         }

```

apps/cli/src/commands/doctor.ts
endersonmaia: That's why it's getting into the `exitCode === 125` branch.

If using `--format`, we may need to add a check for a minimum build version.

My version is this:

```
❯ docker buildx version
github.com/docker/buildx v0.14.0 171fcbe
```

apps/cli/src/commands/doctor.ts
omidasadpour: Already checked and I've got the same error: 

```
    Error: Command failed with ENOENT: docker compose version --short
    spawn docker ENOENT
    Code: ENOENT

```

apps/cli/src/commands/doctor.ts
omidasadpour: Since the old version does not effect on `sunodo`, I would suggest to remove the `--format` from `buildx` check.

endersonmaia: Acctually, I think using at least 0.13.0 would be great for better stability in processing the output.

omidasadpour: Sent a fixup to check the buildx version .

apps/cli/src/commands/doctor.ts
endersonmaia: Actually, I'm using the `true` return in the last commit, where I check if it's ok before checking the `compose` and `buildx` plugin commands.

omidasadpour (approved):

PR #475, opened 2024-04-23
Move of SelfHostedApplicationFactory to @cartesi/rollups

tuler:
More changes of moving SelfHostedApplicationFactory to @cartesi/rollups

Also removes more "sunodo" terms.


endersonmaia (approved):
LGTM

Should we worry about the Socket security report, or can that be handled on another PR before release?

tuler:
> LGTM
> 
> Should we worry about the Socket security report, or can that be handled on another PR before release?

I'll handle the Socket security report when we get to stability of the transition

PR #476, opened 2024-04-23
fixup! refactor(cli): build command

tuler:


endersonmaia (approved):
I'm in favor of keeping using class methods for consistency between calls to `sdkRun`.

But feel free to merge this.

tuler:
> I'm in favor of keeping using class methods for consistency between calls to `sdkRun`.

So you can remove the `async` and make it static

PR #478, opened 2024-04-23
Move token contracts to devnet

tuler:
This moves the token contracts from the token package to the devnet package.
These contracts are for devnet only anyway, so it makes sense.
It will allow us to delete the @sunodo/token package completely, and help with migration to Cartesi CLI.
Previous modifications also made this easier because no other package in the monorepo depends on `@sunodo/token` anymore.

Future PRs will:
- delete the token package
- configure the npm publishing of the devnet package
- install the devnet package in the sdk

### DIFF
--- a/.changeset/good-falcons-work.md
+++ b/.changeset/good-falcons-work.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/devnet": minor
+---
+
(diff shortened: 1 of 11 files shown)


endersonmaia (approved):
LGTM

PR #479, opened 2024-04-23
devnet publishing

tuler:
This PR configures the CI to publish a @sunodo/devnet npm package.
The package will be used by the sdk in a future PR to install it inside the sdk Docker image.
The Docker image of the devnet itself will be deprecated after that.

### DIFF
--- a/.github/workflows/devnet.yaml
+++ b/.github/workflows/devnet.yaml
@@ -12,6 +12,8 @@ on:
                 required: true
             NPM_TOKEN:
                 required: true
+            TURBO_TOKEN:
+                required: true
     pull_request:
         paths:
             - .github/workflows/devnet.yaml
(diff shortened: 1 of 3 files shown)


.github/workflows/devnet.yaml
```diff
@@ -40,6 +46,13 @@ jobs:
             - name: Build
               run: pnpm build --filter @sunodo/devnet
 
+            - name: Publish
+              if: ${{ inputs.release }}
+              run: pnpm publish --access public
+              working-directory: packages/devnet
+              env:
+                  NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
```
endersonmaia, line 60: Won't this conflict with https://github.com/sunodo/sunodo/pull/479/commits/04d3d1e7d17c9b822a29cfbf1ed123fad98b45cd#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR44 ?

tuler: Why?

endersonmaia: Nevermind, I was thinking that it would be possible that the `publish` was going to be executed again.

PR #480, opened 2024-04-23
devnet in SDK

tuler:
This PR adds to the SDK files from the @sunodo/devnet npm package that are necessary to boot an anvil with the Cartesi and test contracts.

Only two files are necessary:
1. `anvil_state.json`: state file used to load anvil with deployed contracts
2. `localhost.json`: information about deployed contracts (ABIs and addresses)

The `localhost.json` is actually used only to print-out information during anvil startup.
I included a `devnet` shell script that was previously the `entrypoint.sh` file of the `sunodo/devnet` Docker image (which will be deprecated).
The command to start the devnet from the sdk image is a little unhandy:

```shell
docker run -e ANVIL_IP_ADDR=0.0.0.0 -p 8545:8545  --entrypoint /usr/local/bin/devnet sunodo/sdk:devel anvil --load-state /usr/share/cartesi/anvil_state.json
```

Maybe we should provide an easier way to do that.
I often use the devnet docker image as a very fast and convenient way to start an anvil with Cartesi by simply using the command:

```shell
docker run -e ANVIL_IP_ADDR=0.0.0.0 -p 8545:8545 sunodo/devnet:1.5.0
```

I published the `@sunodo/devnet@1.5.0` manually to test this. Unfortunately the 1.5.0 had a bug in the publishing that it was not including the `anvil_state.json` file in the package.
That bug is fixed in PR #479 . And there is a commit in this PR that must be deleted before merging.

### DIFF
--- a/.changeset/many-zoos-move.md
+++ b/.changeset/many-zoos-move.md
@@ -1,0 +1,5 @@
+---
+"@sunodo/sdk": minor
+---
+
+add devnet files
--- a/packages/sdk/Dockerfile
+++ b/packages/sdk/Dockerfile
@@ -100,6 +100,7 @@ RUN curl -sSL https://github.com/foundry-rs/foundry/releases/download/nightly-${
     tar -zx -C /usr/local/bin
 
 # healthcheck script using net_listening JSON-RPC method
+COPY devnet /usr/local/bin
 COPY eth_isready /usr/local/bin
 COPY eth_dump /usr/local/bin
 COPY eth_load /usr/local/bin
--- a/packages/sdk/Dockerfile
+++ b/packages/sdk/Dockerfile
@@ -46,6 +46,11 @@ curl -sSL https://github.com/google/go-containerregistry/releases/download/v${CR
     tar -zx -C /usr/local/bin
 EOF
 
+# devnet files
+FROM node:slim as devnet
+ARG DEVNET_VERSION
+RUN npm install -g @sunodo/devnet@${DEVNET_VERSION}
+
 # sdk image
 FROM $SERVER_MANAGER_REGISTRY/$SERVER_MANAGER_ORG/server-manager:$SERVER_MANAGER_VERSION
 ARG SERVER_MANAGER_REGISTRY
@@ -102,6 +107,8 @@ COPY eth_load /usr/local/bin
 COPY entrypoint.sh /usr/local/bin/
 COPY --from=su-exec /usr/local/src/su-exec /usr/local/bin/
 COPY --from=crane /usr/local/bin/crane /usr/local/bin/
+COPY --from=devnet /usr/local/lib/node_modules/@sunodo/devnet/export/abi/localhost.json /usr/share/cartesi/
+COPY --from=devnet /usr/local/lib/node_modules/@sunodo/devnet/build/anvil_state.json /usr/share/cartesi/
 RUN mkdir -p /tmp/.sunodo && chmod 1777 /tmp/.sunodo
 
 ADD --chmod=644 \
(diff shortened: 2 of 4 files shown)


endersonmaia:
> Maybe we should provide an easier way to do that.

We could craft a special `entrypoint.sh` for the SDK with a `devnet`  command.


Callind the `sunodo/sdk` container image like this would start a local anvil instance like before.

```shell
docker run -e ANVIL_IP_ADDR=0.0.0.0 -p 8545:8545 sunodo/sdk:1.5.0 devnet
```

tuler:
> We could ~craft a special~ edit our `entrypoint.sh` for the SDK with a `devnet` command.

> ```shell
> docker run -e ANVIL_IP_ADDR=0.0.0.0 -p 8545:8545 sunodo/sdk:1.5.0 devnet
> ```

The current sdk entrypoint is only to do su-exec.
As you said before I don't know if that is still necessary.

Regarding the devnet "entrypoint", does it need to be a docker entrypoint?
Or is it enough to be a shell script in the `/usr/local/bin` that runs anvil with the required params?
It's also often the case the users need to specify alternative arguments to anvil, like `--block-time`

endersonmaia:
Didn´t test:

```diff
diff --git a/packages/sdk/entrypoint.sh b/packages/sdk/entrypoint.sh
index 7a13e6d..85d3970 100755
--- a/packages/sdk/entrypoint.sh
+++ b/packages/sdk/entrypoint.sh
@@ -1,5 +1,18 @@
 #!/bin/bash
+set -e
 
+# devnet entrypoint
+if [ "$1" = 'devnet' ]; then
+  if [ -z "$ANVIL_IP_ADDR" ]; then
+    echo "You should define ANVIL_IP_ADDR=0.0.0.0 and publish ports."
+    echo "usage: docker run -e ANVIL_IP_ADDR=0.0.0.0 -p 8545:8545 sunodo/sdk devner"
+    exit 1
+  fi
+  shift
+  exec anvil "$@"
+fi
+
+# global enptrypoint
 if [ -z "$GID" -o -z "$UID" -o -z "$USER" -o -z "$GROUP" ]; then
     exec "$@"
 else

```

tuler:
> diff --git a/packages/sdk/entrypoint.sh b/packages/sdk/entrypoint.sh

We don't need an entrypoint, right? Just a simple well crafted script.

endersonmaia:
> > diff --git a/packages/sdk/entrypoint.sh b/packages/sdk/entrypoint.sh
> 
> We don't need an entrypoint, right? Just a simple well crafted script.

IDK, `exec` will ensure `anvil` will be the `pid 1` inside the container and receive/pass the signals instead of depending on a shell.

tuler:
> `exec` will ensure `anvil` will be the `pid 1` inside the container

Is that important?

endersonmaia:
> > `exec` will ensure `anvil` will be the `pid 1` inside the container
> 
> Is that important?

I think so. 

And when we're not sure a program we want to run inside containers implements proper signaling handling, we can use `docker run --init` to do this work.

tuler:
> I think so.

How about an explicit entrypoint to be used with `--entrypoint`?

endersonmaia:
> > I think so.
> 
> How about an explicit alternative entrypoint to be used with `--entrypoint`?
> 
> ```
> docker run --entrypoint devnet (...)
> ```

There should only be a single default `ENTRYPOINT`, and you change based on the `CMD`, that you pass with `docker run <image> CMD`, that's the expected behavior of a container image.

Changing via  `--entrypoint` if for when you want to bypass the default entrypoint.

endersonmaia (approved):
LGTM, remember to squash fixups.

endersonmaia (approved):

PR #481, opened 2024-04-24
Add exec sub-command

omidasadpour:
/close #470 

Example: 

```shell
sunodo exec /bin/bash
```

image

### DIFF
--- a/apps/cli/src/commands/exec.ts
+++ b/apps/cli/src/commands/exec.ts
@@ -4,14 +4,18 @@ import fs from "fs-extra";
 import { lookpath } from "lookpath";
 import path from "path";
 
-export default class Shell extends Command {
-    static description = "Start a shell in cartesi machine of application";
+export default class Exec extends Command {
+    static description = "Execute command in cartesi machine of application";
 
-    static examples = ["<%= config.bin %> <%= command.id %>"];
+    static examples = ["<%= config.bin %> <%= command.id %> /bin/bash"];
 
     static args = {
         image: Args.string({
             description: "image ID|name",
+            required: false,
+        }),
+        command: Args.string({
+            description: "Command to execute",
             required: false,
         }),
     };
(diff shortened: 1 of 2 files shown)


omidasadpour:
> Take a look [here](https://oclif.io/docs/args) on how to handle var args, and not require a `-c` option

Fixed . 
Just I don't know how to allow multiple part commands. like `ls -la`

I have to set it like this : 

```shell
sunodo create "ls -la"
```

apps/cli/src/commands/exec.ts
```diff
@@ -7,32 +7,31 @@ import path from "path";
...
 
-    static examples = ["<%= config.bin %> <%= command.id %>"];
+    static examples = ["<%= config.bin %> <%= command.id %> /bin/bash"];
 
     static args = {
         image: Args.string({
             description: "image ID|name",
             required: false,
         }),
+        command: Args.string({
```
tuler, line 17: What is the image arg above

tuler:
> I just don't know how to allow multiple part commands. like `ls -la`

Did you try the `static strict = false`, as described in the doc?

omidasadpour:
> Did you try the `static strict = false`, as described in the doc?

Yup, Did not work. 
I'll test again

apps/cli/src/commands/exec.ts
omidasadpour: I think we can set a custom image name with it

omidasadpour:
> > Did you try the `static strict = false`, as described in the doc?
> 
> Yup, Did not work. I'll test again

not working

omidasadpour:
> > Take a look [here](https://oclif.io/docs/args) on how to handle var args, and not require a `-c` option
> 
> Fixed . I just don't know how to allow multiple part commands. like `ls -la`
> 
> I have to set it like this :
> 
> ```shell
> sunodo exec "ls -la"
> ```

Actually it does not work because of `-` in. For example I can run this command : 

```
sunodo exec echo $PWD
```

image

omidasadpour:
> > Take a look [here](https://oclif.io/docs/args) on how to handle var args, and not require a `-c` option
> 
> Fixed . I just don't know how to allow multiple part commands. like `ls -la`
> 
> I have to set it like this :
> 
> ```shell
> sunodo exec "ls -la"
> ```

Actually it does not work because of `-`. For example I can run this command without double quotes: 

```shell
sunodo exec ls /home
```

tuler:
Rediscuss the solution for #470 , or if it's necessary

PR #462, opened 2024-04-12
Cartesi CLI

tuler:
Transition from Sunodo CLI to Cartesi CLI.

The goal is to create a branch (a fork) that will be pushed as `main` to the `cartesi/cli` repo.
(I don't intend to merge this PR).

This PR deletes everything that is related to Sunodo. And changes some packages from `sunodo` to `cartesi`.

PR #487, opened 2024-04-24
Remove some sunodo references

tuler:
This removes some references to `sunodo`.
It centralizes for example references to the `.sunodo` directory.

This will facilitate #462

### DIFF
--- a/apps/cli/src/baseCommand.ts
+++ b/apps/cli/src/baseCommand.ts
@@ -57,9 +57,13 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
         return ps?.State;
     }
 
+    protected getContextPath(...paths: string[]): string {
+        return path.join(".sunodo", ...paths);
+    }
+
     protected getMachineHash(): Hash | undefined {
         // read hash of the cartesi machine snapshot, if one exists
-        const hashPath = path.join(".sunodo", "image", "hash");
+        const hashPath = this.getContextPath("image", "hash");
         if (fs.existsSync(hashPath)) {
             const hash = fs.readFileSync(hashPath).toString("hex");
             if (isHash(`0x${hash}`)) {
--- a/apps/cli/src/commands/build.ts
+++ b/apps/cli/src/commands/build.ts
@@ -2,7 +2,6 @@ import { Flags } from "@oclif/core";
 import bytes from "bytes";
 import { execa } from "execa";
 import fs from "fs-extra";
-import path from "path";
 import semver from "semver";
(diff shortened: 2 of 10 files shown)


omidasadpour (commented):
I think it would be better to not use type `any` in https://github.com/sunodo/sunodo/pull/487/files#diff-0eda9573885c4c62dc779278a8b0e8dd493db82e70837b1886e71e09062b8a80R150


This one may make sense instead: 

```code
            if (
                e instanceof Error &&
                (e as Error & { exitCode?: number }).exitCode !== 130
            ) {

```

apps/cli/src/commands/build.ts
```diff
@@ -299,23 +299,23 @@ Update your application Dockerfile using one of the templates at https://github.
...
 
             // create machine snapshot
             if (!flags["skip-snapshot"]) {
                 await this.sdkRun(
                     sdkImage,
                     BuildApplication.createMachineSnapshotCommand(imageInfo),
-                    SUNODO_DEFAULT_EXT2_PATH,
-                    SUNODO_DEFAULT_MACHINE_SNAPSHOT_PATH,
+                    ext2Path,
+                    this.getContextPath("image"),
```
endersonmaia, line 312: For consistency, adding a `const snapshotPath = this.getContextPath("image");` at line 268 and then:

```suggestion
                    snapshotPath,
```

apps/cli/src/baseCommand.ts
```diff
@@ -57,9 +57,13 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
         return ps?.State;
     }
 
+    protected getContextPath(...paths: string[]): string {
+        return path.join(".sunodo", ...paths);
```
endersonmaia, line 61: Could we get the name dynamically like you did wth `${this.config.bin}`, don't know if it is worth but could remove `.sunodo` reference here also.

endersonmaia (changes requested):

apps/cli/src/baseCommand.ts
tuler: I think it's enough to just have this single point with `.sunodo`.
Maybe we will have to support both `.sunodo` and `.cartesi` in the transition. Subject for another PR anyway.

tuler:
>             if (
>                 e instanceof Error &&
>                 (e as Error & { exitCode?: number }).exitCode !== 130
>             ) {
> ```

But it's not that better. You are injecting an optional `exitCode` in an object you don't know anyway.

PR #491, opened 2024-04-26
Deprecate anvil and devnet Docker images

tuler:
Anvil and devnet are now part of SDK 0.6, so we don't have to support independent packages.


endersonmaia (approved):

endersonmaia:
Which milestone?

PR #509, opened 2024-10-29
Upgrade to next 15

tuler:
Upgraded to next 15 and latest mantine.

Need help finding out a hydration issue at `http://localhost:3000/deploy`

```
pnpm i
cd apps/web
pnpm run dev
```


brunomenezes (approved):
@tuler, It looks good. These hydration problems, as far as they're only logs and not breaking the page, should be ok. I am adding a screenshot here where the complaint is because my Grammarly browser plugin injects some classes in the HTML, causing a diff between what comes from the server (SSR) and what is in the client. Once I turn off the plugin, I can refresh as much as I want, and no logs will be displayed regarding this class diff.


nextjs-grammarly-hydration-problem

PR #512, opened 2025-03-14
Alchemy rpc provider

tuler:
Use alchemy as RPC provider.
Followed instructions at https://www.rainbowkit.com/docs/installation#preparing-to-deploy
Created an alchemy app, and will configure at server provider.


endersonmaia (dismissed):
Should we update docs somewhere informing about `ALCHEMY_API_KEY` environment variable ?

